PR #3, opened 2021-06-21
Add codeql analysis

MirahImage:


ChunyiLyu:
Great！

MirahImage:
The error is that the analysis has never been run on the main branch before, so it doesn't have any results to compare with. Once the PR is merged, it will run the analysis on the main branch, and then on all future commits and PRs.

ChunyiLyu (approved):
Thank you!

PR #4, opened 2021-06-27
Exporting Channel State

eibrunorodrigues:
Hey Guys, this is an old PR from streadway, and since we have this "official" fork, I'd still think is a good idea (and I need this haha)

I needed to access the Channel state on a project that I'm participating in. So, I just made the "closed" field public on channel.go in the same way as it is on connection.go. Nothing too fancy, but I think it is important to get the actual status of the channel.

Something that we have on RabbitMQ's client on Python (pika) and .Net (RabbitMQ) and I missed here.

Hope this helps,
Best regards. 

[Original Pull Request](https://github.com/streadway/amqp/pull/486)

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -432,6 +432,12 @@ func (ch *Channel) Close() error {
 	)
 }
 
+// IsClosed returns true if the channel is marked as closed, otherwise false
+// is returned.
+func (ch *Channel) IsClosed() bool {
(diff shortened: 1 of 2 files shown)


michaelklishin (approved):

michaelklishin:
Thank you!

PR #20, opened 2021-10-01
Added support for cert-only login without user and password

mihaitodor:
This PR contains the changes proposed in https://github.com/streadway/amqp/pull/121 together with the changes that @wolf-james requested and some small comment additions for consistency.

Note that I'm not really familiar with this protocol and I haven't looked into setting up a test environment for this setup. People have confirmed that the changes are correct in the original PR, but please let me know if you'd prefer some extra validation and I'll look into it when I'll have some time.

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -54,8 +54,23 @@ func (auth *AMQPlainAuth) Response() string {
 	return buf.String()[4:]
 }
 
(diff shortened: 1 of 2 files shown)


Zerpet:
This PR makes sense to me. WDYT @michaelklishin @ChunyiLyu ?

PR #26, opened 2021-11-22
Fix Signed 8-bit headers being treated as unsigned

alex-goodisman:
Currently the byte tag `'b'` maps to the Go type `byte` which is an alias for `uint8`. This replaces the `'b'` mapping with a `'B'` mapping, the type for unsigned 8-bit values. A new mapping is added for `'b'` that maps to go the Go type `int8`.

Fixes https://github.com/streadway/amqp/issues/391


michaelklishin:
Note: this is consistent with the [RabbitMQ's Errata](https://www.rabbitmq.com/amqp-0-9-1-errata.html) and historically adopted type prefixes.

michaelklishin (approved):
We should ship this with a minor version bump as it is a tiny and subtle, yet breaking change for interop with older versions of this client.

PR #34, opened 2022-01-13
Client example: Use instance logger instead of global logger

andygrunwald:
The `[example_client_test.go](./example_client_test.go)` is using two types of logging mechanism:

https://github.com/rabbitmq/amqp091-go/blob/21ccfdea51148da4de39d35975fdd44480e979e2/example_client_test.go#L89

and

https://github.com/rabbitmq/amqp091-go/blob/21ccfdea51148da4de39d35975fdd44480e979e2/example_client_test.go#L217

This Pull Request is changing the logging behavior to only use the logging instance inside the session struct

### DIFF
--- a/example_client_test.go
+++ b/example_client_test.go
@@ -86,12 +86,12 @@ func New(name string, addr string) *Session {
 func (session *Session) handleReconnect(addr string) {
 	for {
 		session.isReady = false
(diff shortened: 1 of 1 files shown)


MirahImage (approved):

MirahImage:
Thanks!

PR #35, opened 2022-01-13
Client Example: Fix closing order

andygrunwald:
In the client example, the order on how to close elements of the session is wrong:

https://github.com/rabbitmq/amqp091-go/blob/21ccfdea51148da4de39d35975fdd44480e979e2/example_client_test.go#L277-L292

If you first close the channel and after consider the session as done, the reconnection logic will try to re-create the channel and re-init the queue (in the example).
if you first close the done channel, you indicate that we are done with the session and the reconnection logic will end.

### DIFF
--- a/example_client_test.go
+++ b/example_client_test.go
@@ -278,6 +278,7 @@ func (session *Session) Close() error {
 	if !session.isReady {
 		return errAlreadyClosed
 	}
+	close(session.done)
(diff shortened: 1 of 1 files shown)


michaelklishin:
This example really suffers from a newly invented concept, "session", and the confusion between AMQP 0-9-1 channels and Go channels :(

PR #37, opened 2022-01-21
Client example: Renamed concept "Session" to "Client"

andygrunwald:
The concept of a session does not exist in AMQP/RabbitMQ.
This client connect example names the library wrapper "session".
The name can be confusing because it is not usually known in the AMQP/RabbitMQ env or Go env.

This was also raised by @michaelklishin in https://github.com/rabbitmq/amqp091-go/pull/35#issuecomment-1017501288


Zerpet:
Thank you!

Issue #46, opened 2022-02-23
Possible deadlock on DeferredConfirmation.Wait()


ismurov:
Hi, there. 👋

I'm starting to migrate from __github.com/streadway/amqp__ package and delve into the new features of this package. Currently I want to use the __PublishWithDeferredConfirm__ method and it doesn't look production-ready to me. The case in which the problem occurs is closing the connection while waiting for publication confirmation from the AMQP server. 

What I mean in more detail:
The type __Channel__ has a __shutdown__ method in which resources are cleared and closed, also here calls a [__Close__](https://github.com/rabbitmq/amqp091-go/blob/21ccfdea51148da4de39d35975fdd44480e979e2/confirms.go#L102) method of __confirms__ type. But this method does not properly clean up its resources such as __deferredConfirmations__ an awaiting __DeferredConfirmation__.
```go
dc, err := c.PublishWithDeferredConfirm(exchange, key, false, false, msg)
if err != nil {
	log.Fatalf("c.PublishWithDeferredConfirm: %s", err)
}
dc.Wait() // <- deadlock
```

Possible solution:
Add one more method in type __deferredConfirmations__ and call it from __confirms.Close__.
```go
// Close closes all awaiting DeferredConfirmation with Nack.
func (d *deferredConfirmations) Close() error {
	d.m.Lock()
	defer d.m.Unlock()

	for k, v := range d.confirmations {
		v.confirmation = Confirmation{DeliveryTag: k, Ack: false}
		v.wg.Done()
		delete(d.confirmations, k)
	}

	return nil
}
```
But it seems to me that such a situation should be handled with explicit error like ErrClosed. And it's breaking change of package API.

Package version: __v1.3.0__

shifengbin:
the same reason [#44 ](https://github.com/rabbitmq/amqp091-go/pull/44)

PR #47, opened 2022-03-17
Fix deadlock on DeferredConfirmations

SpencerTorres:
This PR improves the changes proposed in #44 originally made by @shifengbin

Changes:
- Fixed deadlock that occurs with DeferredConfirmations when connection is reset
- Made the fix's code more consistent with the other methods in the file
- Added some documentation for `Wait()` and other internal methods
- Added unit test for `deferredConfirmations.Close()`
- Added integration test to confirm that `.Wait()` will return `ack: false` when connection is closed (as suggested by @DanielePalaia )

The newly added unit & integration tests both passed consistently when I ran them

This problem is also mentioned in #46

### DIFF
--- a/confirms.go
+++ b/confirms.go
@@ -103,11 +103,20 @@ func (c *confirms) Close() error {
 	c.m.Lock()
 	defer c.m.Unlock()
 
+	c.releaseWaitConfirms()
+
 	for _, l := range c.listeners {
 		close(l)
 	}
 	c.listeners = nil
 	return nil
+}
+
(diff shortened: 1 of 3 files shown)


SpencerTorres:
It passed for both scenarios interestingly enough. Updated it anyway to publish to an exchange that does not exist

lukebakken (approved):
Thank you!

DanielePalaia (approved):
Looks good! Approved!

PR #54, opened 2022-03-24
changing defaultVersion to buildVersion and create a simple change_ve…

DanielePalaia:
changing defaultVersioning to buildVersioning and made change_version.sh in order to manage versioning


Gsantomaggio (approved):
Ok as the first step. At some point, we have to automate the release process by changing the version based on tag.

MirahImage:
> At some point, we have to automate the release process by changing the version based on tag.
It might be possible to do this with a github action

Issue #31, opened 2021-12-15
Potential race condition in Connection module


Zerpet:
Running our tests with `-race` flag, it reports a race condition in the Connection module. Pasting here one stack trace to investigate.

```
﻿WARNING: DATA RACE
Write at 0x00c00022a050 by goroutine 22:
  github.com/rabbitmq/amqp091-go.(*Connection).shutdown.func1()
      /Users/acedres/workspace/amqp091-go/connection.go:443 +0x50e
  sync.(*Once).doSlow()
      /usr/local/Cellar/go/1.17.3/libexec/src/sync/once.go:68 +0x127
  sync.(*Once).Do()
      /usr/local/Cellar/go/1.17.3/libexec/src/sync/once.go:59 +0x46
  github.com/rabbitmq/amqp091-go.(*Connection).shutdown()
      /Users/acedres/workspace/amqp091-go/connection.go:407 +0x7d
  github.com/rabbitmq/amqp091-go.(*Connection).reader()
      /Users/acedres/workspace/amqp091-go/connection.go:542 +0x344
  github.com/rabbitmq/amqp091-go.Open·dwrap·20()
      /Users/acedres/workspace/amqp091-go/connection.go:250 +0x58

Previous write at 0x00c00022a050 by goroutine 20:
  github.com/rabbitmq/amqp091-go.(*Connection).openComplete()
      /Users/acedres/workspace/amqp091-go/connection.go:847 +0x13d
  github.com/rabbitmq/amqp091-go.(*Connection).openVhost()
      /Users/acedres/workspace/amqp091-go/connection.go:832 +0x1ce
=== RUN   TestChannelOpen
  github.com/rabbitmq/amqp091-go.(*Connection).openTune()
      /Users/acedres/workspace/amqp091-go/connection.go:818 +0xb04
  github.com/rabbitmq/amqp091-go.(*Connection).openStart()
      /Users/acedres/workspace/amqp091-go/connection.go:754 +0x4e7
  github.com/rabbitmq/amqp091-go.(*Connection).open()
      /Users/acedres/workspace/amqp091-go/connection.go:726 +0xc4
  github.com/rabbitmq/amqp091-go.Open()
      /Users/acedres/workspace/amqp091-go/connection.go:251 +0x664
  github.com/rabbitmq/amqp091-go.TestOpen()
      /Users/acedres/workspace/amqp091-go/client_test.go:275 +0x267
  testing.tRunner()
[...]

PR #55, opened 2022-03-25
Test race condition

DanielePalaia:
Some unit tests in connection were showing a race condition (see https://github.com/rabbitmq/amqp091-go/issues/31):
    
These ones were the ones testing Open scenarios. The issue is that Open and Close, rwc.Open and rwc.Close can at the same time write on:
    
    connection.go line 444 and
    connection.go line 849
    

DanielePalaia:
Will be replaced by a new one

PR #56, opened 2022-03-25
Some unit tests in connection were showing a race condition (see #31):

DanielePalaia:
These ones were the ones testing Open scenarios. The issue is that Open and Close, rwc.Open and rwc.Close can at the same time write on:

connection.go line 444 and
connection.go line 849

While it's not clear if the library should protect this eventuality, the tests are testing the Open function, so I think the close can be put in the main thread avoiding the race and not affecting the test validity

### DIFF
--- a/change_version.sh
+++ b/change_version.sh
@@ -1,0 +1,4 @@
+#/bin/bash
+echo $1 > VERSION
+sed -i -e "s/.*buildVersion           = \"*.*/buildVersion =            \"$1\"/" ./connection.go
+go fmt ./...
--- a/client_test.go
+++ b/client_test.go
@@ -236,7 +236,7 @@ func TestDefaultClientProperties(t *testing.T) {
 	if want, got := defaultLocale, srv.start.Locale; want != got {
 		t.Errorf("expected locale %s got: %s", want, got)
 	}
-	rwc.Close()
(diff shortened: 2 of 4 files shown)


Zerpet (changes requested):
The problem with this approach is that `t.Fatalf()` stops the execution of the function, and in that case, `Close()` will never be called, potentially leaving open connections behind.

I would place the `Close()` call in a [t.Cleanup()](https://pkg.go.dev/testing#T.Cleanup) call to ensure it is called, regardless of the result of the test.

DanielePalaia:
@Zerpet good idea put it in all TestOpen tests

Zerpet:
We should call `t.Cleanup()` at the begining of the test, as soon as possible, not at the end 😄 

I'm not sure about including [f781c65](https://github.com/rabbitmq/amqp091-go/pull/56/commits/f781c65b0fac4bd1347941b3f1f5ca56ba91c937) and [9bfc51b](https://github.com/rabbitmq/amqp091-go/pull/56/commits/9bfc51b6dda5c27990d3a6bf749df000476eacc9) in this PR. I agree those are useful, but I don't see how they are relevant to the race condition 🤔

DanielePalaia:
Superseded by https://github.com/rabbitmq/amqp091-go/pull/57

PR #57, opened 2022-03-25
These ones were the ones testing Open scenarios. The issue is that Op…

DanielePalaia:
…en and Close, rwc.Open and rwc.Close can at the same time write on:

connection.go line 444 and
connection.go line 849

While it's not clear if the library should protect this eventuality, the tests are testing the Open function, so I think the close can be put in the main thread avoiding the race and not affecting the test validity

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -218,7 +218,7 @@ func TestDefaultClientProperties(t *testing.T) {
 
 	go func() {
 		srv.connectionOpen()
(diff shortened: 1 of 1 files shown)


Gsantomaggio (approved):

DanielePalaia:
> Thank you!

Thanks to you! @Zerpet @Gsantomaggio

PR #58, opened 2022-03-28
bump versioning to 1.3.2

DanielePalaia:


DanielePalaia:
This PR is including also some fixes autogenerated by go fmt ./... as go fmt is called by the change_version.sh script

Gsantomaggio (approved):

PR #59, opened 2022-03-30
adding documentation on NotifyClose of connection and channel to enfo…

DanielePalaia:
…rce consuming returned channels (see PR issue_18)


Gsantomaggio (approved):
per conversation with @DanielePalaia we will merge this PR and we will add a specific section to the documentation about deadlocks

PR #66, opened 2022-04-06
Regenerate certs with SHA256 to fix test with Go 1.18+

anthonyfok:
This fixes "x509: cannot verify signature: insecure algorithm SHA1-RSA"
error with Go 1.18 and up in TestTLSHandshake.

Thanks to @lnussbaum for the bug report; see https://bugs.debian.org/1008423

(See streadway/amqp#525 for the equivalent pull request at the original repository.)


Zerpet (approved):
Thank you!

PR #74, opened 2022-04-19
ci: run tests with race detector

fho:
```
ci: enable the race detector when running go tests

Pass "-race" when running the tests via the Makefile or in CI.
This enables the race detector and can catch some bugs.

make: define all targets as phony

This prevents that Make skips running a target if a file with the same
name then target exist.
```


lukebakken (approved):

fho:
@lukebakken I now realized that project also has a github actions config which is used.
So "-race" also needs to be added when running tests via github actions.

Is travis CI also used somehow? Or could the config be removed?

PR #95, opened 2022-06-29
Added function to get the remote peer's IP address (conn.RemoteAddr())

rabb1t:
Hi, 

In the clustered RabbitMQ installation it is often necessary to get the remote node's IP address to be able to determine to which node exactly you are connected to.

Having that in mind, I've added `conn.RemoteAddr()` method which behaves similar to `conn.LocalAddr()` but returns remote peer IP address:port instead of local one.

Thank you and have a good day!

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -265,6 +265,18 @@ func (c *Connection) LocalAddr() net.Addr {
 	return &net.TCPAddr{}
 }
 
+/*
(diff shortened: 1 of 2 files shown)


Zerpet (approved):

Zerpet:
I will merge after the checks pass successfully.

PR #101, opened 2022-07-13
fix race condition with context close and confirm at the same time on DeferredConfirmation.

sapk:
When doing more testing, I found one race condition is now possible with the PR #96.

This add test to detect the race condition and add basic solution to prevent to happen.

Sorry for this little possible bug.

One better way would be to use atomic with a uint32 as a bool but that imply potentially a lot of code change


michaelklishin (approved):

lukebakken (approved):
Good catch! Thanks!

PR #114, opened 2022-09-01
Add support for connection.update-secret

Zerpet:
This method is in the AMQP 0.9.1 spec and it is supported by RabbitMQ.
This is useful when the secret used in the connection has an expiry, as
it allows the applications to renew the secret without the hassle of
closing-reconnecting consumers/producers.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -266,6 +266,12 @@ func Open(conn io.ReadWriteCloser, config Config) (*Connection, error) {
 	return c, c.open(config)
 }
 
+/*
+UpdateSecret updates the secret used to authenticate this connection. It is used when
+secrets have an expiration date and need to be renewed, like OAuth 2 tokens.
+
+It returns an error if the operation is not successful, or if the connection is closed.
+*/
 func (c *Connection) UpdateSecret(newSecret, reason string) error {
 	if c.IsClosed() {
 		return ErrClosed
--- a/connection.go
+++ b/connection.go
@@ -266,6 +266,16 @@ func Open(conn io.ReadWriteCloser, config Config) (*Connection, error) {
 	return c, c.open(config)
 }
 
+func (c *Connection) UpdateSecret(newSecret, reason string) error {
+	if c.IsClosed() {
+		return ErrClosed
+	}
+	return c.call(&connectionUpdateSecret{
+		NewSecret: newSecret,
+		Reason:    reason,
+	}, &connectionUpdateSecretOk{})
+}
+
 /*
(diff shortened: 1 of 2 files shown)


Gsantomaggio:
Looks to me.
I tested it with https://github.com/rabbitmq/rabbitmq-oauth2-tutorial 

```golang

package main

import (
	"context"
	"encoding/json"
	"fmt"
	amqp "github.com/rabbitmq/amqp091-go"
	"io/ioutil"
	"net/http"
	"strings"
	"time"
)

func main() {
	fmt.Println(GetToken())

	connection, err := amqp.DialConfig(fmt.Sprintf(
		"%s://%s:%s@%s:%s/%s",
		"amqp",
		"",
		GetToken(),
		"localhost",
		"5672",
		"/",
	), amqp.Config{})

	if err != nil {
		panic(err)
	}
	channel, err := connection.Channel()
	if err != nil {
		panic(err)
	}
	for i := 0; i < 1000; i++ {

		err := channel.PublishWithContext(context.Background(), "", "test-key", false, false, amqp.Publishing{})
		if err != nil {
			fmt.Printf("Error %s\n", err)
			return
		}

		if i%59 == 0 {
			err := connection.UpdateSecret(GetToken(), "test-key")
			fmt.Printf("Update secret\n")
			if err != nil {
				fmt.Printf("UpdateSecret Error %s\n", err)

				return
			}
		}

		time.Sleep(1 * time.Second)
		fmt.Printf("Sent %d\n", i)
	}
	defer connection.Close()

}

type Access struct {
	AccessToken string `json:"access_token"`
}

func GetToken() string {
	url := "http://localhost:8080/realms/test/protocol/openid-connect/token"
	method := "POST"

	payload := strings.NewReader("client_id=producer&client_secret=kbOFBXI9tANgKUq8vXHLhT6YhbivgXxn&grant_type=client_credentials")

	client := &http.Client{}
	req, err := http.NewRequest(method, url, payload)

	if err != nil {
		fmt.Println(err)
		return ""
	}
	req.Header.Add("Content-Type", "application/x-www-form-urlencoded")

	res, err := client.Do(req)
	if err != nil {
		fmt.Println(err)
		return ""
	}
	defer res.Body.Close()

	body, err := ioutil.ReadAll(res.Body)
	if err != nil {
		fmt.Println(err)
		return ""
	}

	var x = Access{}
	err = json.Unmarshal(body, &x)
	if err != nil {
		return err.Error()
	}
	return x.AccessToken
}


```

Gsantomaggio (approved):

PR #118, opened 2022-09-26
remove publishing msgs and closing attempts on already closed channels.

wishwaprabodha:
We are using the amqp library in many production applications in our Company. Suddenly we realised that after a random TCP timeout from the RMQ server consumers were not attached. This connection drop happened to NodeJS, RoR and Golang applications and only Golang applications failed to reconnect. After figuring out what happened, We try to stimulate the Issue by disconnecting and reconnecting the VPN. After reconnection, it keeps returning a panic error as we are trying to publish to a channel which is already closed. This change fixed the issue we had but we can't use the change since only fix we could do this from the library end. Please look into this. thanks

Screenshot from 2022-09-19 09-54-47

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -454,25 +454,8 @@ func (c *Connection) shutdown(err *Error) {
 		c.m.Lock()
 		defer c.m.Unlock()
 
-		if err != nil {
-			for _, c := range c.closes {
-				c <- err
-			}
-		}
-
-		if err != nil {
-			c.errors <- err
-		}
-		// Shutdown handler goroutine can still receive the result.
-		close(c.errors)
-
-		for _, c := range c.closes {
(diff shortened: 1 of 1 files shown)


Zerpet (changes requested):
This PR is consistently blocking the test `TestOpenFailedCredentials`. On top of that, just removing part of the shutdown logic is doubtly a solution. I'd be inclined to approach this with a try-and-recover mechanism, similar to what's described [here](https://go.dev/blog/defer-panic-and-recover).

Also to note that the screenshot shows you were using streadway/amqp, instead of this library. I assume you reproduced successfully the issue using this library at some point?

michaelklishin:
No activity in a month => closing.

PR #142, opened 2022-12-19
Fix inefficient use of buffers that reduces the potential throughput of basicPublish

fadams:
As per issue https://github.com/rabbitmq/amqp091-go/issues/141

Improve the performance of Channel sendOpen method by adding an sendUnflushed method to Connection that allows us to use the write buffer more efficiently by writing all the Frames of the message and *then* flushing the buffer rather than flushing each Frame. This significantly improves the performance of basicPublish for small messages where the bulk of the CPU load tends towards Syscall so with small messages we might have a Syscall per message vice per Frame.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -447,14 +447,20 @@ func (c *Connection) send(f frame) error {
 	return err
 }
 
-// The send() method uses writer.WriteFrame(), which will write the Frame then
+// sendUnflushed performs an *Unflushed* write. It is otherwise equivalent to
+// send(), and we provide a separate flush() function to explicitly flush the
+// buffer after all Frames are written.
+//
+// Why is this a thing?
+//
+// send() method uses writer.WriteFrame(), which will write the Frame then
 // flush the buffer. For cases like the sendOpen() method on Channel, which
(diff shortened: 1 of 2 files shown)


connection.go
```diff
@@ -447,6 +447,69 @@ func (c *Connection) send(f frame) error {
...
+		})
+	} /*else { // Moving to flush increases basicPublish throughput even more
+		// Broadcast we sent a frame, reducing heartbeats, only
+		// if there is something that can receive - like a non-reentrant
+		// call or if the heartbeater isn't running
+		select {
+		case c.sends <- time.Now():
+		default:
+		}
+	}*/
```
Zerpet, line 482: Let's remove this commented code, since it made it to `flush` and the comment explaining the reasoning is also there.

Zerpet (commented):
Thank you for contributing to this project. This looks good to me 👍  There's commented code that can be removed. I'm going to QA this locally before I give the approval ✅

connection.go
fadams: Howdy. FWIW I retained the commented out code in the pull request so that the logic of why it was done might be clearer to whomever was reviewing, I was totally expecting/intending for it to be removed IDC.

PR #145, opened 2022-12-22
Add constants for Queue arguments

Zerpet:
Queues accept optional arguments during queue declaration. Most of those
arguments can be set using a policy (recommended), however, queue type
must be set at queue declaration. To pave the way for Quorum and Stream
queue adoption, this commit adds constants to easily set the queue type.

Other common used arguments are part of this commit. Additional queue
arguments can be added as needed/requested by the community.


lukebakken (approved):
I made one small change. I realize it's not necessary but it caught me by surprise that it worked (https://go.dev/ref/spec#Constant_declarations)

PR #155, opened 2023-01-18
Remove outdated information from README

scriptcoded:
Removes outdated information in README as requested and according to https://github.com/rabbitmq/amqp091-go/issues/154.


lukebakken (approved):
Thanks!

PR #169, opened 2023-02-02
adding go 1.20 to tests

halilylm:
I added go version 1.20 to testing workflow.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        go-version: ['1.18', '1.19', '1.20']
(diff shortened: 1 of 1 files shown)


lukebakken (approved):
Thank you!

slagiewka:
@lukebakken @halilylm if this project is supposed to follow the two latest releases the `setup-go` action supports `stable` and `oldstable` versions to be used instead of regular tags.

Zerpet:
We try to keep up to date with new versions of Go, and ensure they are compatible. At the same time, we don't want to break existing users unless we really have to. That's one of the reasons we still have some `x interface{}` instead of `x any` 🙂 That being said, I like the idea of switching to `stable` and `oldstable` tags.

PR #183, opened 2023-03-17
Fix race condition on confirms

calloway-jacob:
This PR fixes Issue #182

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -1437,7 +1437,7 @@ func (ch *Channel) PublishWithDeferredConfirmWithContext(ctx context.Context, ex
 
 	var dc *DeferredConfirmation
 	if ch.confirming {
-		dc = ch.confirms.Publish()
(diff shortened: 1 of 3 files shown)


lukebakken (approved):
@Zerpet changing `Publish` to `publish` is an API change, but I consider it being public to be a bug in the first place. What do you think?

Zerpet (approved):

Zerpet:
> @Zerpet changing `Publish` to `publish` is an API change, but I consider it being public to be a bug in the first place. What do you think?

I'm happy with that. `confirm` is unexported struct, so it's not even a breaking change, since user code can't use it.

Zerpet:
Thank you @calloway-jacob !! 🚀

calloway-jacob:
@Zerpet and @lukebakken Very happy to help!

PR #186, opened 2023-03-22
Fix Example Client not reconnecting

### LINKED ISSUE #185: Example reconnecting Client sometimes won't reconnect
### Describe the bug

I have a client written using your Example Client Test using the reconnection logic in there that I have been testing and hit an issue where it would not reconnect, but would get stuck when trying to create a new Channel on a broken Connection.

This code is the problem:

https://github.com/rabbitmq/amqp091-go/blob/f706086fb623dff2fa1ca62e912d9ac0527b71c1/example_client_test.go#L199-L210

### Reproduction steps

I found this by complete accident.

My app is running on a box that has a wired ethernet connection and a 4G backup. I was testing failing over to the 4G backup by physically pulling the ethernet cable from the box. And then reconnecting to test failing back to wired.

In doing this several times I just happened to hit the right timing where the wired connection was dropped between the connection being created successfully and the channel creation call being made.

I've no idea how you would automate this test!

### Expected behavior

Example code that demonstrates reconnection should reconnect 😊

### Additional context

This looks like a pretty simple fix, by adding a check of the ```client.notifyConnClose``` in that ```select``` block, I'll create a PR shortly.

frankfil:
Fixes #185

### DIFF
--- a/example_client_test.go
+++ b/example_client_test.go
@@ -204,6 +204,9 @@ func (client *Client) handleReInit(conn *amqp.Connection) bool {
 			select {
 			case <-client.done:
 				return true
+			case <-client.notifyConnClose:
+				client.logger.Println("Connection closed. Reconnecting...")
+				return false
 			case <-time.After(reInitDelay):
 			}
 			continue


Zerpet (approved):

Zerpet:
Thank you!

PR #206, opened 2023-06-21
Constant for consumer timeout queue argument

### LINKED ISSUE #201: Add a constant to set consumer timeout as queue argument
Starting in RabbitMQ 3.12, we can set a consumer timeout via a policy
and via a queue argument. The queue argument is `x-consumer-timeout`.

https://rabbitmq.com/consumers.html#acknowledgement-timeout

rabbitmq/rabbitmq-server/issues/5437

Zerpet:
Documentation to queue argument constants was missing links to the
constant themselves. Go Doc allows to link to a specific symbol using
squared brackets.

Added as well a constant for SAC in CQ and QQ.

Fixes #201

### DIFF
--- a/tls_test.go
+++ b/tls_test.go
@@ -66,7 +66,7 @@ func (s *tlsServer) Serve(t *testing.T) {
 }
 
 func startTLSServer(t *testing.T, cfg *tls.Config) tlsServer {
-	l, err := tls.Listen("tcp", "127.0.0.1:0", cfg)
+	l, err := tls.Listen("tcp", "127.0.0.1:3456", cfg)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


tls_test.go
```diff
@@ -66,7 +66,7 @@ func (s *tlsServer) Serve(t *testing.T) {
 }
 
 func startTLSServer(t *testing.T, cfg *tls.Config) tlsServer {
-	l, err := tls.Listen("tcp", "127.0.0.1:0", cfg)
+	l, err := tls.Listen("tcp", "127.0.0.1:3456", cfg)
```
Zerpet, line 69: I'm not sure if this is necessary, but it seems to resolve flakes with tests in Windows container.

PR #221, opened 2023-09-23
Update write.go

pinkfish:
Change the time comparison to use Equal, which is safer with timestamps in go.


lukebakken (approved):
Thank you!

michaelklishin (approved):

PR #223, opened 2023-09-27
Closes 222

yywing:
Closes https://github.com/rabbitmq/amqp091-go/issues/222

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -813,13 +813,16 @@ func (c *Connection) allocateChannel() (*Channel, error) {
 
 // releaseChannel removes a channel from the registry as the final part of the
 // channel lifecycle
-func (c *Connection) releaseChannel(id uint16) {
+func (c *Connection) releaseChannel(ch *Channel) {
 	c.m.Lock()
 	defer c.m.Unlock()
 
 	if !c.IsClosed() {
-		delete(c.channels, id)
-		c.allocator.release(int(id))
+		got, ok := c.channels[ch.id]
+		if ok && got == ch {
+			delete(c.channels, ch.id)
+			c.allocator.release(int(ch.id))
+		}
 	}
 }
 
@@ -831,7 +834,7 @@ func (c *Connection) openChannel() (*Channel, error) {
 	}
 
 	if err := ch.open(); err != nil {
-		c.releaseChannel(ch.id)
+		c.releaseChannel(ch)
 		return nil, err
 	}
 	return ch, nil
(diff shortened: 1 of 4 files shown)


connection.go
```diff
@@ -640,9 +640,12 @@ func (c *Connection) dispatch0(f frame) {
...
 		// lolwat - channel0 only responds to methods and heartbeats
-		if err := c.closeWith(ErrUnexpectedFrame); err != nil {
-			Logger.Printf("error sending connectionCloseOk with ErrUnexpectedFrame, error: %+v", err)
-		}
+		// closeWith use call don't block reader
+		go func() {
+			if err := c.closeWith(ErrUnexpectedFrame); err != nil {
+				Logger.Printf("error sending connectionCloseOk with ErrUnexpectedFrame, error: %+v", err)
+			}
+		}()
```
Zerpet, line 648: This approach has a lot of potential to cause routine leaks. What's the difference between using a go routine and simply synchronously call the function?

See the following test failure what detected a go routine leak: https://github.com/rabbitmq/amqp091-go/actions/runs/6324272238/job/17185854481?pr=223#step:5:333

connection.go
```diff
@@ -689,9 +692,12 @@ func (c *Connection) dispatchClosed(f frame) {
...
 			// unexpected method on closed channel
-			if err := c.closeWith(ErrClosed); err != nil {
-				Logger.Printf("error sending connectionCloseOk with ErrClosed, error: %+v", err)
-			}
+			// closeWith use call don't block reader
+			go func() {
+				if err := c.closeWith(ErrClosed); err != nil {
+					Logger.Printf("error sending connectionCloseOk with ErrClosed, error: %+v", err)
+				}
+			}()
```
Zerpet, line 700: Same comment as before, I suspect this will leak go routines.

Zerpet (commented):
Thank you for this PR. I agree with most of the changes. See my comments below for outstanding issues.

connection.go
yywing: https://github.com/rabbitmq/amqp091-go/issues/222#issuecomment-1736798083

> n some case, reader will call c.closeWith, and block at c.call at <- c.rpc. But only reader write to <-c.rpc - -

`dispatch0` and `dispatchN` is uesed at `reader` function. If we sync call this fucniton, `reader` will block.

yywing: fixed

Zerpet: I'm happy with the changes in f4ea4b2b288a3863365f021f45a429099e305086

Zerpet:
Thank you!

PR #224, opened 2023-09-27
Update spec091.go

pinkfish:
Fix up a go lint error with no space between the // and the text.

### DIFF
--- a/gen.ps1
+++ b/gen.ps1
@@ -1,0 +1,14 @@
+$DebugPreference = 'Continue'
+$ErrorActionPreference = 'Stop'
+
+Set-PSDebug -Off
+Set-StrictMode -Version 'Latest' -ErrorAction 'Stop' -Verbose
+
+New-Variable -Name curdir  -Option Constant -Value $PSScriptRoot
+
+$specDir = Resolve-Path -LiteralPath (Join-Path -Path $curdir -ChildPath 'spec')
+$amqpSpecXml = Resolve-Path -LiteralPath (Join-Path -Path $specDir -ChildPath 'amqp0-9-1.stripped.extended.xml')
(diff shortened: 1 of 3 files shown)


lukebakken:
@Zerpet is this file auto-generated? I will see if I can figure that out.

@pinkfish we do go lint checking as part of GitHub actions. How was this formatting missed?

pinkfish:
I think there is more strict lint checking in some areas :)

lukebakken:
> I think there is more strict lint checking in some areas :)

@pinkfish - That's good to know, but I'm interested in _specific details_:
* Which linter are you using?
* How can I run it to see this particular formatting issue being raised as an error?

As @Zerpet said, as far as our current linter is concerned, there is no issue here, so we'll need more information.

lukebakken (changes requested):
The code that generates `spec091.go` must be updated, and re-run.

pinkfish:
I think it is go vet buildtag

It is because the no space after a // should be a compiler tag for the +build pieces.

lukebakken:
@pinkfish hmm this is what I get when I run that command:

```
C:\Users\lbakken\development\rabbitmq\amqp091-go [main ≡]
> go vet buildtag
package buildtag is not in std (C:\Users\lbakken\scoop\apps\go\current\src\buildtag)
```

It seems like `-buildtag` is the correct argument. But, when I run it, no errors are returned:

```
C:\Users\lbakken\development\rabbitmq\amqp091-go [main ≡]
> go vet -all
C:\Users\lbakken\development\rabbitmq\amqp091-go [main ≡]
> go vet -buildtag
```

lukebakken (approved):

PR #232, opened 2023-11-29
Documenting Publishing.Expiration usage

niksteff:
So this one got me good on a really nasty bug from my side. I was using the pkg to publish big batches of messages to a queue. When reading the source code it was not clear to me the value was used as milliseconds. I assumed wrongly if the value is `0` the message will not expire. I assumed this because `0` would be the nil value for a go int. This assumption was painfully wrong - which lead to all messages not immediately retrieved by the consumer to get rightfully deleted. I wanted to spare others from the same mistake.

I am perfectly aware the rabbit mq documentation states the usage here: https://www.rabbitmq.com/ttl.html#per-message-ttl-in-publishers

But I think a little more documentation in the code and some clarification can prevent others from making the same mistake. I also provided two constants for the field that further elaborate on the usage and should explain it in its entirety. They are not needed but so are the `DeliveryMode` constants (Transient, Persistent). If this is not wanted please provide feedback and I will refine the PR.

Best regards and thank you all a lot for your work on this client!

### DIFF
--- a/types.go
+++ b/types.go
@@ -144,6 +144,19 @@ const (
 	flagReserved1       = 0x0004
 )
 
+// Expiration. These constants should be used to set a messages expiration TTL.
+// They should be viewed as a clarification of the expiration functionality in
+// messages and their usage is not enforced by this pkg.
+//
+// The server requires a string value that is interpreted as milliseconds. If
+// no value is set, which translates to the nil value of string, the message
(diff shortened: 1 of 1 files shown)


Zerpet (approved):
Thank you!

Zerpet:
I will merge after CI passes since we have a restriction in place for merging 🔒

PR #233, opened 2023-12-12
Add notify middleware to notify functions in channel

n25a:
I've added `NotifyMiddleware` to all notify functions in this PR.
One of the uses of this middleware is that we can easily handle rabbitmq reconnection.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -505,7 +505,7 @@ graceful close, no error will be sent.
 In case of a non graceful close the error will be notified synchronously by the library
 so that it will be necessary to consume the Channel from the caller in order to avoid deadlocks
 */
-func (ch *Channel) NotifyClose(c chan *Error) chan *Error {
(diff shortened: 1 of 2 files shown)


lukebakken (changes requested):
Thanks for using this library. We can't accept a PR without tests, or documentation that shows _why_ this feature is being added, and what problem it is solving.

But, before you spend more time on it, my initial reaction is that this PR will not be accepted - @Zerpet ?

Zerpet:
> Thanks for using this library. We can't accept a PR without tests, or documentation that shows _why_ this feature is being added, and what problem it is solving.
> 
> But, before you spend more time on it, my initial reaction is that this PR will not be accepted - @Zerpet ?

I agree. I had a quick look and I don't understand the intent, nor what's incomplete or missing in the current implementation. Definitely a PR with no tests and no justification for the lack of tests is a no-no.

Zerpet:
Closing due to staleness

PR #248, opened 2024-02-27
Increasing the code coverage

edercarloscosta:
This PR aims to add two simple tests:

- auth.go
- types.go

The main goal is to increase the code coverage on two opportunities I saw taking into consideration that doesn't add noises or need to change on the main code, only tests.


lukebakken (approved):
Thank you!

PR #250, opened 2024-02-28
Example fix

Boris-Plato:
Related to https://github.com/rabbitmq/amqp091-go/issues/249

### DIFF
--- a/_examples/pubsub/pubsub.go
+++ b/_examples/pubsub/pubsub.go
@@ -91,11 +91,12 @@ func publish(sessions chan chan session, messages <-chan message) {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
+	pending := make(chan message, 1)
+
 	for session := range sessions {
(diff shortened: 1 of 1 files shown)


_examples/pubsub/pubsub.go
```diff
@@ -117,6 +118,7 @@ func publish(sessions chan chan session, messages <-chan message) {
 			select {
 			case confirmed, ok := <-confirm:
 				if !ok {
+					pub.Close()
```
Gsantomaggio, line 121: Won't it be closed anyway here? https://github.com/rabbitmq/amqp091-go/blob/main/_examples/pubsub/pubsub.go#L135

Zerpet: I don't think so, because the `break Publish` exits the `for` loop.

Zerpet (commented):
It's looking good. I'm going to verify manually before merging.

Zerpet (approved):
Thank you!

PR #256, opened 2024-04-09
Connection/Channel.NotifyClose: allow only buffered channel=1

ikavgo:
see https://github.com/knative-extensions/eventing-rabbitmq/issues/1369


Zerpet (changes requested):
Do we want to force the buffer size to be 1? What's wrong with higher buffer sizes?

I had a in mind a check to ensure that we don't use unbuffered channels i.e. `cap(ch) != 0`. What's the reasoning to enforce the capacity to always be one?

I'm unsure if we want to `panic` either, it seems a bit extreme to crash the entire program just because the argument of a library function was incorrect. I think we should look for alternatives to introduce this change, for example:

- Create a new function that returns a channel and an error, and return a non-nil error if channel is buffered; and deprecate this function.
- Use a channel of capacity 1 if an unbuffered channel is passed (bit ugly), else use the channel as-is.
- Any other approach than panicking is welcome.

PR #257, opened 2024-04-23
Update Go directive

Zerpet:
In #251, we used `URI.Has()` function, which was introduced in Go 1.17.
We also have a dependency on uber-go-leak, which requires Go 1.20. In
addition, we only test in "oldstable" and "stable" versions in CI. That
means, we test in version N and N-1, where N is the latest available
version of Go.

Attempting to build the tests in Go 1.16 errors with:

```
# github.com/rabbitmq/amqp091-go [github.com/rabbitmq/amqp091-go.test]
./uri.go:148:11: params.Has undefined (type url.Values has no field or method Has)
./uri.go:156:11: params.Has undefined (type url.Values has no field or method Has)
./uri.go:164:11: params.Has undefined (type url.Values has no field or method Has)
```

Attempting to build with 1.17 yields:

```
gmake tests
go test -race -v -tags integration
# go.uber.org/goleak/internal/stack
../../go/pkg/mod/go.uber.org/goleak@v1.3.0/internal/stack/stacks.go:123:19: undefined: errors.Join
../../go/pkg/mod/go.uber.org/goleak@v1.3.0/internal/stack/stacks.go:256:18: undefined: strings.CutPrefix
note: module requires Go 1.20
FAIL	github.com/rabbitmq/amqp091-go [build failed]
gmake: *** [Makefile:20: tests] Error 2
```


lukebakken (approved):
Great!

PR #280, opened 2024-08-20
Fixing simple errors

korolev-d-l:
- Fixed channel max error message
- Fixed example "set client connection name"


michaelklishin:
The `golangci-lint` failure is equally present in `main` and seems to be new (possibly due to a linter update?)

korolev-d-l:
> The `golangci-lint` failure is equally present in `main` and seems to be new (possibly due to a linter update?)

Yes, my old version 1.59.1 dont showes error, but after update golangci-lint to 1.60.2 it showed

michaelklishin:
@Zerpet any objections to merging this?

Zerpet (approved):
Happy to merge this PR.

PR #286, opened 2024-10-29
Feat: defined a type for exchange kinds

techerfan:
The developer must type out the kind of exchange when declaring it, which could lead to typos. Also, the IDE or code editor did not help or suggest a type of exchange.
This way, selecting an exchange is more straightforward and helps maintain the code more easily.

### DIFF
--- a/_examples/pubsub/pubsub.go
+++ b/_examples/pubsub/pubsub.go
@@ -68,7 +68,7 @@ func redial(ctx context.Context, url string) chan chan session {
 				log.Fatalf("cannot create channel: %v", err)
 			}
 
-			if err := ch.ExchangeDeclare(exchange, "fanout", false, true, false, false, nil); err != nil {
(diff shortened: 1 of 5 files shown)


Zerpet:
I agree with the idea, but the current implementation changes the public API, and it's a breaking change. We can merge this PR when we plan to release 2.0.

In the mean time, you can accomplish the same without declaring the `ExchangeType`, simply declaring the `string` constants. If you can submit a different PR (keep this one open please), I'll gladly merge a non-breaking change.

techerfan:
It's ok. I'll keep the PR open.

> In the mean time, you can accomplish the same without declaring the `ExchangeType`, simply declaring the `string` constants. If you can submit a different PR (keep this one open please), I'll gladly merge a non-breaking change.

These string constants were [defined](https://github.com/rabbitmq/amqp091-go/blob/main/types.go#L19) before. I just changed their name and added `ExchangeType` to them.

PR #294, opened 2025-01-16
Add warning about concurrency with Channels

### LINKED ISSUE #291: consume input basicConsumeOk but response queueBindOk
hi,
When I used the channel. constame() method, the service returned 503 and I logged out directly, but I found that the server actually showed that the binding was successful. I forked version v1.10 and added printing, but found that it was a type mismatch. The type passed in is basicConsumeOk, but the returned type is queueBindOk. This will cause the client to exit but the server still experiences message loss due to this consumption.

image
I am using AWS's EKS cluster and AWS's RabbiMT version 3.13.6

Zerpet:
This fixes #291.

It is tribal knowledge that channels should not be used accessed by
multiple goroutines. Adding this warning to the function documentation
should avoid further issues like #291.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -897,6 +897,10 @@ func (c *Connection) closeChannel(ch *Channel, e *Error) {
 Channel opens a unique, concurrent server channel to process the bulk of AMQP
 messages.  Any error from methods on this receiver will render the receiver
 invalid and a new Channel should be opened.
+
+Channels are not thread-safe. To avoid unexpected behavior, do not share
+a single Channel instance between multiple goroutines. Concurrent calls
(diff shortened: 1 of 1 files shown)


lukebakken (approved):

snehanshu-socialmobile:
So what's the solution? One channel per goroutine? Or a dedicated goroutine for interacting with the channel?

psnehanshu:
So what's the solution? One channel per goroutine? Or a dedicated goroutine for interacting with the channel?

Zerpet:
Either solution is fine, depending on your needs. If you have further questions, I'd recommend starting a discussion here: https://github.com/rabbitmq/amqp091-go/discussions

PR #307, opened 2025-11-19
Change the error type for ErrDeliveryNotInitialized

kl09:
We should catch the error that Channel/Connection is closed and reconnect. Better to catch the error by the same code: ChannelError.

### DIFF
--- a/delivery.go
+++ b/delivery.go
@@ -6,11 +6,8 @@
 package amqp091
 
 import (
-	"errors"
 	"time"
 )
-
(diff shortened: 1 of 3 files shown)


Zerpet (changes requested):
We simply cannot accept a PR that has reduced the test coverage without justification.

Why is better the new error type?
What use case does cover that wasn't covered by the previous sentinel error?

michaelklishin:
Our team unanimously agreed that we are not interested in accepting this change.

PR #330, opened 2026-03-31
 fix: respect context cancellation on publishing with context operations

NawafSwe:
# PR Description:                                                                                                                                                                                                                                                                                                                                                                                                                                                         
  This PR addresses an issue with the publishing interface where:                                                                                                                                                                        
  - PublishWithContext                                                                                                                                                                                                                   
  - PublishWithDeferredConfirmWithContext                                                                                                                                                                                                                                                                                                                                                                                                                                
  
They aren't honoring context when passed.                                                                                                                                                                                                                                                                                                                                                                                                                                       
## Changes                                                                                                                                                                                                                                                                                                                                                                                                                                                        
  - Run the underlying publish call in a goroutine and use select to race it against ctx.Done(), allowing the caller to return early on context cancellation or timeout                                                                  
  - Updated doc comments to clarify that on context cancellation, the publish may still complete in the background since context is not propagated to the underlying Publish call                                                        
  - Added unit tests for pre-cancelled context and deadline exceeded scenarios   


This PR was opened to address #329

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -1492,8 +1492,10 @@ func (ch *Channel) Publish(exchange, key string, mandatory, immediate bool, msg 
 /*
 PublishWithContext sends a Publishing from the client to an exchange on the server.
 
-NOTE: If the context is cancelled, the publish operation may still complete in the background because the context is not propagated to the underlying Publish call.
-In this case, PublishWithContext will return the context error immediately rather than waiting for Publish to finish.
+If the context is already cancelled when PublishWithContext is called, it
+returns the context error immediately without attempting to publish.  Context
+cancellation after the call has started does not interrupt an in-flight
+Publish, as the underlying I/O is not context-aware.
 
 When you want a single message to be delivered to a single queue, you can
 publish to the default exchange with the routingKey of the queue name.  This is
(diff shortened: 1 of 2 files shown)


NawafSwe:
Hello @lukebakken appreicate your review

lukebakken:
@NawafSwe **DO NOT EVER** ping maintainers to get their attention. We all have day jobs where we're _paid_ to do work that is higher priority.

Zerpet (changes requested):
We can't take this approach to honour context cancellation. More details following soon.

Zerpet:
The approach on this PR does not honour context cancellation either. By spawning a go routine to publish, and then racing on context cancellation or `Publish` return, **without cancelling the loser**, you aren't honouring the context.

One situation where this approach surprises:

RabbitMQ is blocked due to memory pressure -> `Publish()` is called -> go routine blocks in the background -> context times out and "wins" the `select` -> `Publish` returns an error.
Next second, RabbitMQ unblocks, no longer memory pressure -> go routine in the background completes the call to `PublishWithDeferredConfirm()` -> message is published to RabbitMQ.

Now you have a message in RabbitMQ, and an error returned from `Publish...()`. Context was not honoured because the call to Publish was not cancelled. This is not what we want. I would argue is counter productive even.

The most we can accept in v1 is a gate keeper check at the beginning of the function, something like:

```go
select {
  case <-ctx.Done():
    return ctx.Err()
  default:
}
```

So that if the context is already cancelled when `Publish...()` is called, the function immediately returns, without even trying to publish to RabbitMQ.

NawafSwe:
@Zerpet, thanks for sharing the details.
I will address this concern next

lukebakken (changes requested):
> If the context is cancelled, the publish operation may still complete in the background because the context is not propagated to the underlying Publish call

As @Zerpet stated, this is unacceptable.

NawafSwe:
I have updated the PR with the suggested changes

Zerpet (approved):
Thank you, this looks good. I'll merge after CI passes.

Zerpet:
hey @lukebakken, any objection to merge this PR with new changes?

Sorry for the direct mention, I'd like to release a new version soon-ish, since our last release was in May 2024 🙁

lukebakken (approved):

lukebakken:
No worries, feel free to merge anytime you approve something 😸 

Obviously this is the bare-bones use of context. Cancellation is a real PITA, ask me how I know from the .NET client...

PR #339, opened 2026-06-01
Feature: implement automatic connection and channel recovery with state change notifications

suchitd:
## Proposed Changes

Introduces automatic reconnection and channel recovery capabilities, allowing client
applications to transparently recover from connection and channel closures caused by 
network failures.
Key Features:
- **Auto-Recovery**: Automatic connection re-dial and re-establishment of all channels,
  publisher confirmation settings, and active consumer subscriptions. The recovery configuration/interface  
  can be provided by the consuming application, otherwise, the default configuration/interface implementation will be considered.
- **NotifyStateChange**: Added a lifecycle state machine to both `Connection` and `Channel` 
  broadcasting transitions (`open`, `reconnecting`, `closing`, `closed`) to registered 
  listeners. This allows applications to cleanly block activity during recovery. The states and Lifecycle type
  definition is used as is from [rabbitmq-amqp-go-client](https://github.com/rabbitmq/rabbitmq-amqp-go-client/blob/main/pkg/rabbitmqamqp/life_cycle.go)
- **Integration Test (`recovery_test.go`)**: Included integration tests 
  that use the RabbitMQ management plugin HTTP API to forcefully drop connections
  (HTTP utils is used as is from [rabbitmq-amqp-go-client](https://github.com/rabbitmq/rabbitmq-amqp-go-client/blob/main/pkg/test-helper/http_utils.go)) and verify:
  - Clean state transitions flow through `NotifyStateChange`
  - Active consumer subscriptions are restored, and unacknowledged messages are safely 
    redelivered (with `Redelivered: true`) and successfully acknowledged post-recovery
- **Example Usage (`_examples/recovery/recovery.go`)**: Added the recovery example to
  demonstrate how applications can listen to `NotifyStateChange` events and use synchronizing
  primitives (like `sync.Cond`) to gracefully block publishers while the connection is 
  reconnecting and unblock them once the state returns to open.

## Types of Changes

- [ ] Bugfix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

## Further Comments

Topology Recovery will be implemented as a follow-up change. With this commit alone, the producer and consumer will work with auto-recovery as long as queues are durable.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -540,6 +540,9 @@ func (ch *Channel) IsClosed() bool {
 }
 
 // NotifyStateChange registers a listener for state changes.
+//
+// It is necessary to continuously consume from the channel passed to NotifyStateChange
+// to avoid blocking internal state dispatch routines and leaking goroutines.
 func (ch *Channel) NotifyStateChange(c chan *StateChanged) {
 	ch.lifeCycle.notifyStateChange(c)
 }
--- a/channel.go
+++ b/channel.go
@@ -142,7 +142,7 @@ func (ch *Channel) shutdown(e *Error) {
 		}
 	}
 
-	if e == nil || !ch.connection.IsRecoveryEnabled() {
+	if e == nil || !ch.connection.IsRecoveryEnabled() || !ch.connection.isRecoverable(e) {
 		ch.consumers.close()
 
 		for _, c := range ch.closes {
--- a/channel.go
+++ b/channel.go
@@ -7,9 +7,11 @@ package amqp091
 
 import (
 	"context"
+	"math/rand"
 	"reflect"
 	"sync"
 	"sync/atomic"
+	"time"
 )
 
 // 0      1         3             7                  size+7 size+8
@@ -26,10 +28,11 @@ exchange.  Errors on methods with this Channel as a receiver means this channel
 should be discarded and a new channel established.
 */
 type Channel struct {
-	destructor sync.Once
-	m          sync.Mutex // struct field mutex
-	confirmM   sync.Mutex // publisher confirms state mutex
-	notifyM    sync.RWMutex
+	destructorM sync.Mutex
+	destructed  bool
+	m           sync.Mutex // struct field mutex
+	confirmM    sync.Mutex // publisher confirms state mutex
+	notifyM     sync.RWMutex
 
 	connection *Connection
 
@@ -74,6 +77,8 @@ type Channel struct {
 	message messageWithContent
 	header  *headerFrame
 	body    []byte
+
+	reconnecting sync.Mutex
 }
 
 // Constructs a new channel with the given framing rules
@@ -100,23 +105,34 @@ func (ch *Channel) setClosed() {
 func (ch *Channel) shutdown(e *Error) {
 	ch.setClosed()
 
-	ch.destructor.Do(func() {
-		ch.m.Lock()
-		defer ch.m.Unlock()
+	ch.destructorM.Lock()
+	if ch.destructed {
+		ch.destructorM.Unlock()
+		return
+	}
+	ch.destructed = true
+	ch.destructorM.Unlock()
 
-		// Grab an exclusive lock for the notify channels
-		ch.notifyM.Lock()
-		defer ch.notifyM.Unlock()
+	ch.m.Lock()
+	defer ch.m.Unlock()
 
-		// Broadcast abnormal shutdown
-		if e != nil {
-			for _, c := range ch.closes {
-				c <- e
-			}
-			// Notify RPC if we're selecting
-			ch.errors <- e
+	// Grab an exclusive lock for the notify channels
+	ch.notifyM.Lock()
+	defer ch.notifyM.Unlock()
+
+	// Broadcast abnormal shutdown
+	if e != nil {
+		for _, c := range ch.closes {
+			c <- e
 		}
+		// Notify RPC if we're selecting
+		select {
+		case ch.errors <- e:
+		default:
+		}
+	}
 
+	if e == nil || !ch.connection.IsRecoveryEnabled() {
 		ch.consumers.close()
 
 		for _, c := range ch.closes {
@@ -149,7 +165,10 @@ func (ch *Channel) shutdown(e *Error) {
 		close(ch.errors)
 		close(ch.close)
 		ch.noNotify = true
-	})
+	} else {
+		close(ch.errors)
+		close(ch.close)
+	}
 }
 
 // send calls Channel.sendOpen() during normal operation.
@@ -1124,7 +1143,16 @@ func (ch *Channel) Consume(queue, consumer string, autoAck, exclusive, noLocal, 
 
 	deliveries := make(chan Delivery)
 
-	ch.consumers.add(consumer, deliveries)
+	config := ConsumerConfig{
+		Queue:     queue,
+		Consumer:  consumer,
+		AutoAck:   autoAck,
+		Exclusive: exclusive,
+		NoLocal:   noLocal,
+		NoWait:    noWait,
+		Args:      args,
+	}
+	ch.consumers.add(consumer, deliveries, config)
 
 	if err := ch.call(req, res); err != nil {
 		ch.consumers.cancel(consumer)
@@ -1228,7 +1256,16 @@ func (ch *Channel) ConsumeWithContext(ctx context.Context, queue, consumer strin
 
 	deliveries := make(chan Delivery)
 
-	ch.consumers.add(consumer, deliveries)
+	config := ConsumerConfig{
+		Queue:     queue,
+		Consumer:  consumer,
+		AutoAck:   autoAck,
+		Exclusive: exclusive,
+		NoLocal:   noLocal,
+		NoWait:    noWait,
+		Args:      args,
+	}
+	ch.consumers.add(consumer, deliveries, config)
 
 	if err := ch.call(req, res); err != nil {
 		ch.consumers.cancel(consumer)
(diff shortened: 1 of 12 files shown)


Gsantomaggio:
I haven't checked the code, but recovery connections work as expected.  

Using `recovery.go` example

lukebakken (commented):
## Review Summary

AI-assisted review (Claude Code, Anthropic) with @lukebakken guidance. Full findings with code references, impact analysis, and suggested fixes: https://gist.github.com/lukebakken/7917947d63f83e2e43d35b9ae8cedc6e

### Critical

1. **Channel ID allocator collision after recovery** - `openTune()` creates a fresh allocator during reconnect, but recovered channels still occupy their old IDs. New channels allocated post-recovery can collide, overwriting recovered channels in the map and causing protocol-level corruption.

2. **Data race in `SetRecoverableErrorCodes` / `AddRecoverableErrorCodes`** - These methods write to `RecoverableErrorCodes` without synchronization while `isRecoverable()` reads the same slice concurrently from recovery goroutines.

3. **`DefaultRecoverableErrorCodes` shared by reference** - When `ReconnectionConfig.RecoverableErrorCodes` is nil, the global slice is assigned directly without copying (line 321), creating shared mutable state between connections and the package-level variable.

### Significant

4. **Fixed retry interval** - Consider exponential backoff with a shorter initial delay (1-2s + larger jitter for thundering-herd spread) growing to a 10-30s ceiling, rather than a fixed 5s on every attempt including the first.

5. **`deliverLoop` goroutine leak** - If the user stops reading from a `NotifyStateChange` channel, the delivery goroutine blocks forever and the internal queue grows without bound.

6. **`NotifyStateChange` replaces previous listener** - Unlike `NotifyClose`/`NotifyBlocked`/etc. which support multiple listeners, this silently replaces the previous one. API inconsistency.

7. **Dead deduplication check** - `lifecycle.go:116` compares interface values with `==`, but every call site allocates a fresh pointer, so the check is always false.

8. **No cancellation for in-progress recovery** - `time.Sleep()` cannot be interrupted. Users cannot abort recovery via `Close()` if it is called before `resetState()` runs.

9. **Latent race: reader's `defer close(c.rpc)`** - Theoretically the old reader's defer could close the new `c.rpc` after `resetState()` replaces it. Not practically exploitable at default 5s intervals, but a correctness concern with short intervals.

### Minor

- `ILifeCycleState` / `IConnectionRecovery` naming not idiomatic Go
- Credentials logged in plain text via `conn.url`
- Zero-value `MaxRetryCount` silently disables recovery
- `DefaultConnectionRecovery.config` field is unused (dead code)
- "continuosly" typo (3 instances in example)
- Test utilities hardcode `localhost:15672` and `guest:guest`

internal/utils/http.go
Zerpet, line 1: I would like this file to either be moved to `internal/` or renamed as a test file `http_test.go`. Otherwise, the types in this file will be import-able by consumers of this library. I understand the types are behind the `integration` build tag, but I don't think that's sufficient.

internal/utils/http.go
```diff
@@ -0,0 +1,105 @@
...
+		return "", err3
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode == 200 { // OK
+		bodyBytes, err2 := io.ReadAll(resp.Body)
+		if err2 != nil {
+			return "", err2
+		}
```
Zerpet, line 90: I'm curious about the choice of using 3 distinct error variables `err`, `err2` and `err3`. It is idiomatic in Go to re-use the same `err` variable, despite shadowing.

confirms.go
```diff
@@ -123,6 +123,21 @@ func (c *confirms) Close() error {
 	return nil
 }
 
+// Reset clears any pending deferred confirmations and resets the sequencer
+// state for recovery, while keeping the listeners intact.
+func (c *confirms) Reset() {
```
Zerpet: Do we need this exported? I believe is only used internally within the package.

connection.go
```diff
@@ -625,44 +749,72 @@ func (c *Connection) flush() (err error) {
...
+
+	if err != nil {
+		for _, listener := range c.closes {
+			select {
+			case listener <- err:
+			default:
+				// If blocked/full, send in a goroutine so we never deadlock the shutdown sequence
+				go func(l chan *Error, e *Error) {
+					l <- e
+				}(listener, err)
```
Zerpet: This can leak go routines if the receiver never reads the channel. I would rather attempt to delivery and skip if the channel is full, like:

```go
select {
case l <- e:
// ok
default:
// skip
}

File: lifecycle.go
Comment (Zerpet) [inline lifecycle.go:1]: I'm not a fan of the approach for `ILifeCycleState`. My first objection is the name, it's borrowed from C#. I don't think we need the interface, or at least not an exported interface.

I would refactor so that we use a new type, based on `byte`, with a Stringer implementation and exported constants for different lifecycle states. Like so (incomplete, just to showcase the idea):

```go
type StateLifeCycle byte

func (s StateLifeCycle) String() string {
	if s == StateOpen {
		return "open"
	}
	return "closed"
}

const (
	StateOpen = iota
	StateClosed
)
```

Why? Because this saves a type switch to the end user, and the user experience is just comparing a value you get from the library with a constant in the library. That's more elegant than type switching. This also allows to just pass constant values internally, instead of allocating new structs when we send or set a lifecycle state.

lifecycle.go
```diff
@@ -0,0 +1,225 @@
...
+		return "closing"
+	case stateClosed:
+		return "closed"
+	}
+	return "unknown"
+
+}
+
+// StateChanged defines the connection or channel life cycle.
+// See ILifeCycleState for more details about the possible states.
```
Zerpet: Let's just export the constants and document the meaning of each state in the constants.

lifecycle.go
```diff
@@ -0,0 +1,225 @@
...
+	listeners []*stateListener // The registered state change listeners.
+	mutex     *sync.Mutex      // The mutex to protect the state changes.
+}
+
+func NewLifeCycle() *LifeCycle {
+	return &LifeCycle{
+		state: &StateClosed{},
+		mutex: &sync.Mutex{},
+	}
+}
```
Zerpet: Why this needs to be exported?

I'm not sure we are expecting users to create `LifeCycle` structs.

Zerpet (commented):
Sending some comments. PR not reviewed fully, I'll continue tomorrow.

connection.go
suchitd: I agree, however there is a trade-off if we simply skip, we might end up in application-level deadlock iff listener channel is unbuffered and recovery is enabled. 
For e.g., 
1. Application registers unbuffered listener := conn.NotifyClose(make(chan *Error))
2. Perform some action that caused shutdown
3. within some milliseconds: application yet to read from <-listener , before that shutdown skipped sending error on listener channel cause of default case as it was not ready to consume.
4. Now, application is blocked at <-listener

suchitd: A better solution would be to add a timeout so that goroutine exits after timeout:
```
    select {
    case l <- e:
    case <-time.After(5 * time.Second):
        // Give up to avoid leaking the goroutine permanently
    }
```

internal/utils/http.go
suchitd: Done! Moved to `internal` !

internal/utils/http.go
suchitd: Sorry about that, copy-paste error from: https://github.com/rabbitmq/rabbitmq-amqp-go-client/blob/main/pkg/test-helper/http_utils.go

Fixed it now!

confirms.go
suchitd: Good catch! Made it private!

connection.go
suchitd: I've added a timeout for the goroutine, this will take care of not leaking the goroutines, however there is still a chance of missing the error message.

At application level, user must use buffered channel for NotifyClose() and start listening to it immediately. This will prevent creation of additional goroutines.

lifecycle.go
suchitd: You are right, it should be private. Made that change!

I'll revisit these changes and re-evaluate if any members are exported that shouldn’t have been. Thanks!

lifecycle.go
suchitd: Done

lifecycle.go
suchitd: This makes complete sense. Thanks! I've made the suggested changes.

Gsantomaggio: > I'm not a fan of the approach for ILifeCycleState. My first objection is the name, it's borrowed from C#.

That's my fault :)!

connection.go
Zerpet: I would argue that an application dead-locking on receiving from a notification channel is poor design and an app issue. Nonetheless, I like the approach of waiting for 5 seconds before giving up on delivering the notification 👍

lifecycle.go
```diff
@@ -0,0 +1,190 @@
...
+			l.mutex.Unlock()
+			return
+		}
+		sc := listener.queue[0]
+		listener.queue[0] = nil // Allow GC to reclaim the popped element
+		listener.queue = listener.queue[1:]
+		ch := listener.ch
+		l.mutex.Unlock()
+
+		ch <- sc
```
Zerpet, line 139: I'm pondering about this. Generally this is dangerous, because if there's no receiver reading or the channel is full, the function can block indefinitely. Since this function is internal and the `lifeCycle` struct is also internal, we should be in full control of the receiver and could ensure that we always receive. I need to review this a bit more carefully.

What do you think?

channel.go
```diff
@@ -149,7 +181,16 @@ func (ch *Channel) shutdown(e *Error) {
 		close(ch.errors)
 		close(ch.close)
 		ch.noNotify = true
-	})
+
+		var err error
+		if e != nil {
+			err = errors.New(e.Error())
+		}
```
Zerpet, line 191: By doing this, we are losing the error type. We have some sentinel errors in `types.go` that could be used to match error types, for example, using `errors.Is(err, amqp.ErrChannelMax`. By doing `errors.New(e.Error())`, it loses the type `*amqp.Error` and becomes a string error, therefore failing any `errors.Is()` check against our sentinel errors.

You can rewrite this as `err = e` or wrap the error with `err = fmt.Errorf("shutdown error: %w, e)`

connection.go
```diff
@@ -190,6 +204,28 @@ func DialTLS_ExternalAuth(url string, amqps *tls.Config) (*Connection, error) {
...
+func DialRecovery(url string, recovery *Recovery) (*Connection, error) {
+	if recovery == nil {
+		recovery = &Recovery{}
+	}
+	config := Config{
+		Locale:   defaultLocale,
+		Recovery: recovery,
+	}
+	return DialConfig(url, config)
+}
```
Zerpet: I would prefer to not have this new constructor, because we are already suffering a bit of function signature explosion, as we have:

1. Dial
2. DialConfig
3. DialTLS
4. DialTLS_ExternalAuth

Adding a 5th function for exactly the same is not great. I don't mind if you leave it here for now, but I'll deprecate it and trim it in the v2 version of this library 🙃

connection.go
```diff
@@ -383,6 +447,40 @@ func (c *Connection) NotifyClose(receiver chan *Error) chan *Error {
...
+// and closing all recovery cancellation listeners.
+func (c *Connection) CloseRecovery() {
+	c.m.Lock()
+	defer c.m.Unlock()
+
+	for _, listener := range c.recoveryCancels {
+		close(listener)
+	}
+	c.recoveryCancels = nil
+}
```
Zerpet: I think we can un-export this function. Is there a use case to clear to the recovery behaviour while keeping the connection open?

connection.go
```diff
@@ -625,44 +749,76 @@ func (c *Connection) flush() (err error) {
...
+	if err == nil || !c.IsRecoveryEnabled() {
 		c.channels = nil
 		c.allocator = nil
 		c.noNotify = true
-	})
+
+		var e error
+		if err != nil {
+			e = errors.New(err.Error())
+		}
```
Zerpet, line 821: Same as for channel

channel.go
```diff
@@ -517,6 +567,37 @@ func (ch *Channel) NotifyClose(c chan *Error) chan *Error {
...
+// and closing all recovery cancellation listeners.
+func (ch *Channel) CloseRecovery() {
+	ch.m.Lock()
+	defer ch.m.Unlock()
+
+	for _, listener := range ch.recoveryCancels {
+		close(listener)
+	}
+	ch.recoveryCancels = nil
+}
```
Zerpet: Same as for `connection.go`

channel.go
```diff
@@ -1841,3 +1940,197 @@ func (ch *Channel) GetNextPublishSeqNo() uint64 {
...
+
+// watchChannel watches the channel for close events and triggers recovery if needed.
+func (ch *Channel) watchChannel() {
+	errCh := ch.NotifyClose(make(chan *Error, 1))
+	go func() {
+		for err := range errCh {
+			if err != nil {
+				Logger.Printf("Channel %d closed unexpectedly: %v", ch.id, err)
+				if ch.connection.Config.Recovery != nil && ch.connection.Config.Recovery.ConnectionRecovery != nil {
```
Zerpet, line 2000: Perhaps it's a bit more efficient to check if connection recovery is enabled _before_ dispatching the go-routine? And become a no-op if connection recovery is not enabled?

Zerpet (commented):
Another round of comments

channel.go
suchitd: Done!!!

connection.go
suchitd: Done!!

connection.go
suchitd: Yeah, even I was hesitant adding another wrapper method just for Recovery. I am not even using it in integration tests except the recovery example. That also I updated to use `DialConfig`.

However, I am now setting the default locale if its empty in `DialConfig`:
```
	if config.Locale == "" {
		config.Locale = defaultLocale
	}
```

Looking at the documentation of `Locale`, it should always be set. So, I guess it should be fine to set it always if empty. Please let me know otherwise.

connection.go
suchitd: If application user wants to interrupt active recovery without calling connection/channel close right away (considering close will cleanup everything), they could use this method. 
I don't have a concrete usecase at the moment. Maybe we can make this private for now, and see if such need arise in the future.

channel.go
suchitd: The `watchChannel` and `watchConnection` are already guarded by `IsRecoveryEnabled`. So, if recovery is not enabled, it won't register `NotifyClose` and start the goroutine.

connection.go
suchitd: Made it private.

channel.go
suchitd: Made it private. Please refer to my comment on previous thread regarding the same comment on `connection.go`

connection.go
Zerpet: Yeah, setting a default locale is fine.

lifecycle.go
suchitd: This method is designed in such a way that the listener should receive state change events in strictly ordered manner. 
Maybe we can document for NotifyStateChange (just like NotifyClose) that the listener should continue to consume from channel to avoid leaking goroutines.

suchitd: I thought about this for a while, and I think we should not drop state change event (i.e. it should be blocking and ordered). Consider, an application is waiting for state change from `reconnecting` to `open` or `close`, and if we drop that message for some reason, the application will not wake up to know reconnection succeeded and failed. Now, thats an issue from library side. 
Application must always listen from the channel if they are subscribed for NotifyStateChange, if thats not happening, its an application issue.

suchitd: > we should be in full control of the receiver and could ensure that we always receive

The receiver channel comes from application through `NotifyStateChange(ch chan *StateChanged)`. The receiver is not in library control. Am I missing something?

I've added a doc comment on `NotifyStateChange` that the application must always consume from the channel passed to `NotifyStateChange` to make sure it always receives.

Zerpet: We can go as is, and change it later if we receive reports. 👍

Gsantomaggio:
Great :)!

PR #357, opened 2026-06-23
feature: implement automatic topology recovery

suchitd:
## Proposed Changes

- Define the `TopologyRecovery` interface and `DefaultTopologyRecovery` implementation to manage topology restoration.
- Introduce `TopologyConfiguration` and associated configuration structs (`QosConfig`, `ExchangeConfig`, `QueueConfig`, `BindingConfig`, `ExchangeBindingConfig`) to track active topology state on each channel.
- Automatically record QoS settings, exchange/queue declarations, and bindings on the channel when topology recovery is enabled.
- Automatically remove recorded topology upon deletion or unbinding of exchanges and queues, performing in-place cleanup of related bindings to avoid extra allocations.
- Re-declare and re-bind all tracked topology elements during channel setup (`setupChannel`) upon successful reconnection.
- Handle server-generated queue name changes during recovery, automatically updating the internal queue registry, bindings, and active consumer configurations with the new server-assigned names.

## Types of Changes

- [ ] Bugfix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist


Gsantomaggio:
I would prefer the same implementation [we did in AMQP 1.0](https://github.com/rabbitmq/rabbitmq-amqp-go-client/blob/a551c17f24aa8f484e907fb00076648523db39e2/pkg/rabbitmqamqp/amqp_connection.go#L52-L64)  

To give the user the opportunity to decide what to recover. Restoring the persistent queues/exchanges/bindings by default can be a side effect, since someone could decide to delete queues/exchanges/bindings.


By default, the client should restore the transient entities, and then the user can decide what to restore

suchitd:
> I would prefer the same implementation [we did in AMQP 1.0](https://github.com/rabbitmq/rabbitmq-amqp-go-client/blob/a551c17f24aa8f484e907fb00076648523db39e2/pkg/rabbitmqamqp/amqp_connection.go#L52-L64)
> 
> To give the user the opportunity to decide what to recover. Restoring the persistent queues/exchanges/bindings by default can be a side effect, since someone could decide to delete queues/exchanges/bindings.
> 
> By default, the client should restore the transient entities, and then the user can decide what to restore

The user can provide their own implementation of `TopologyRecovery` where they have a complete and more granular control of which entities to recover based on `channel.TopologyConfiguration()`. What do you think?

Gsantomaggio:
I leave Aitor to validate the code, but I approve the [options to define the topology recovery](https://github.com/rabbitmq/amqp091-go/pull/357/changes#diff-25d86b8f6eb8f1dd2f3773a47bd953172e5a465cb5a8acc13402033b82fd8c00R91-R117)

channel.go
```diff
@@ -115,6 +115,65 @@ type Channel struct {
 	lifeCycle    *lifeCycle // The current state of the channel.
 
 	recoveryCancels []chan struct{} // listeners for channel recovery cancellation
+
+	topologyM             sync.Mutex            // Mutex for the topology configuration.
+	topologyConfiguration TopologyConfiguration // The current topology configuration of the channel.
```
suchitd: In AMQP 0-9-1, tracking topology at the channel level may cause recovery failures due to unpredictable channel ordering. Unlike AMQP 1.0's connection-level management, independent AMQP 0-9-1 channels can experience out-of-order recovery failures. For instance, if Channel 2 recovers first and tries to create a binding before Channel 1 has recovered the required exchange and queue, a channel exception will occur and halt the recovery process.

I'll come up with a solution to solve this problem.

suchitd: Moved topology configuration at connection level and recover topology in phases across channels. We first re-establish all the channels and then based on the dependency order(exchanges → queues → bindings → exchange-to-exchange bindings → consumers), we recover exchanges, queues, bindings, and consumers across all the channels.

lukebakken:
Since this PR appears to still be in-progress, I converted it to draft mode. I'm going to do an initial review with the 🧞 today.

lukebakken (changes requested):
Thanks for this work. I reviewed it thoroughly against the Java and .NET clients and ran the recovery suite against a live broker. Found three issues with fixes proposed in #361 (based on this branch), plus a few design questions.

## Blocking

- Binding removal ignores `Args`. `recordBinding` dedups on `(Queue, Exchange, Key, Args)`, but `removeBinding`/`removeExchangeBinding` and the global merge match only `(Queue, Exchange, Key)`. Unbinding one of several arg-distinct bindings (e.g. headers exchange) purges all of them, so survivors are lost on recovery. Fixed in #361 with a regression test.
- Recovery-exhaustion panic. When topology recovery fails on every retry, `cleanup()` double-closes the `NotifyClose`/`Blocked` listeners and panics with "close of closed channel", crashing the process. Fixed in #361 with an integration test.

## Design questions (not blocking, worth discussing)

- Default mode is `TopologyRecoveryOnlyTransient`. Both Java and .NET default to recovering all topology and have no transient/durable distinction. Is the divergence intended?
- A single entity that fails to recover aborts the whole connection recovery and drives it to closed, taking healthy channels with it. Java continues past per-entity failures; .NET continues except on channel-level errors. Consider continuing and surfacing failures.
- No "forget on auto-delete" cascade: an auto-delete queue whose last consumer is cancelled (or auto-delete exchange whose last binding is removed) stays tracked and can be resurrected on recovery. Both reference clients drop these.

## Note

While reviewing I also found a pre-existing data race in `Connection.shutdown` (present on `main`, unrelated to this PR), filed as #360.

Gsantomaggio:
@lukebakken, Thank you for the review. 


> Default mode is TopologyRecoveryOnlyTransient. Both Java and .NET default to recovering all topology and have no transient/durable distinction. Is the divergence intended?

I wanted to add this option and I added the same to the AMQP 1.0 clients. As user I may change the topology in another way, let's say via management UI. The client could be not aligned with the new topology, suppose the client raises the auto-reconnection for some reason ( it can happen after days) the topology changes again due of the topology recovery. 

So, I wanted to give more options to the user to avoid unexpected behaviors. With transient the user can restore only the resources associated to the connection.

suchitd:
Thanks @lukebakken for the review and fixing the identified issues.

> A single entity that fails to recover aborts the whole connection recovery and drives it to closed, taking healthy channels with it. Java continues past per-entity failures; .NET continues except on channel-level errors. Consider continuing and surfacing failures.

The current design philosophy treats recovery as binary: either everything is restored and the application continues seamlessly, or recovery fails and the application handles it explicitly.
Partial recovery—where some entities succeed and others are skipped—creates a hidden middle state. It gives the application a false sense of normalcy while actually degrading its behavior (e.g., missing consumers, dropped bindings).
Even if we surface those skipped entities, if they aren't handled properly at the application layer, it introduces undesired behaviors because the library only handled the job halfway.
While this may differ from the standard Java/.NET approach, a clean failure signal is more honest and actionable than partial success. That said, I am completely open to discussion if we have a use case for partial recovery. What do you think? :)

> No "forget on auto-delete" cascade: an auto-delete queue whose last consumer is cancelled (or auto-delete exchange whose last binding is removed) stays tracked and can be resurrected on recovery. Both reference clients drop these.

Good catch! I’m aware of this gap and completely agree we should handle the 'forget on auto-delete' cascade to match the reference clients.
Since this PR is already growing quite large, would it be alright if we address this in a dedicated follow-up PR? I'll open an issue for it so we can track it properly. Please let me know if its fine with you.

lukebakken:
IMHO all of our official AMQP 0.9.1 clients should behave as closely as possible, barring differences in language idioms. So, I'll re-iterate that I think the decision to only recover transient entities is not the right one, nor is the "all-or-nothing" recovery. Both differ from the Java and .NET clients. I'll leave it at that, though!

Gsantomaggio:
>  I'll re-iterate that I think the decision to only recover transient entities is not the right one

Maybe it is not clear. There is the possibility of recovering the `transient entities`, `all the entities` or `disable` . [It is only a configuration.](https://github.com/rabbitmq/amqp091-go/pull/357/changes#diff-25d86b8f6eb8f1dd2f3773a47bd953172e5a465cb5a8acc13402033b82fd8c00R91-R117). We can change the default if you prefer. But I'd like to leave the possibility to the user to decide. Since we already did it. 

```
TopologyRecoveryOnlyTransient TopologyRecoveryMode = iota
TopologyRecoveryAllEnabled
TopologyRecoveryDisabled
```


so `TopologyRecoveryAllEnabled` is like Java and .NET

suchitd:
Thanks @lukebakken & @Gsantomaggio for the reviews.

Just to summarise, we have following open items:
1. Continue topology recovery on one or more failed entities recovery and surface the failures
2. Forget on auto-delete
3. Recover all entities by default (TopologyRecoveryAllEnabled)

I've already started working on these items and will prepare follow-up PR(s).
I am making above changes as a part of [feature/topology-recovery-enhancements](https://github.com/rabbitmq/amqp091-go/tree/feature/topology-recovery-enhancements) branch.

Gsantomaggio:
Ok with me. Thanks both!

suchitd:
@Zerpet Please feel free to drop your comments in this PR. I'll make sure to address them through follow-up PRs if needed. Thanks!

Issue #360, opened 2026-06-30
Data race in Connection.shutdown between buffered listener send goroutine and close(listener)


lukebakken:
### Summary

`Connection.shutdown` has a data race on the `NotifyClose` listener channels. When a listener's buffer is full, shutdown spawns a goroutine to send on the channel (connection.go:798-808), and the same shutdown invocation may then `close()` that channel (connection.go:821-823). The concurrent send and close race on the channel. This is independent of topology recovery and reproduces on `main`.

### Race report

Observed under `go test -race` against RabbitMQ 4.x:

```
WARNING: DATA RACE
Read at 0x... by goroutine 615:
  runtime.chansend()
  amqp091.(*Connection).shutdown.func1()    connection.go:803
  amqp091.(*Connection).shutdown.gowrap3()   connection.go:807

Previous write at 0x... by goroutine 613:
  runtime.closechan()
  amqp091.(*Connection).shutdown()           connection.go:822
  amqp091.(*Connection).reader()             connection.go:967
```

### Mechanism

1. A `NotifyClose` listener is registered with a small (or zero) buffer and is not drained promptly.
2. A recoverable shutdown fills the buffer (the value is never read).
3. A later non-recoverable shutdown finds the buffer full, so it takes the `default` branch and launches the send goroutine (`go func(){ ... listener <- err ... }()`, guarded by `recover()`).
4. The same shutdown then reaches the non-recoverable branch and calls `close(listener)`.
5. The spawned goroutine's `chansend` races with `closechan`. The existing `recover()` catches the panic from sending on a closed channel, but it does not make the channel-state access race-free, so `-race` reports it and the ordering is not guaranteed safe.

### Impact

A real concurrency bug flagged by the race detector. In practice the `recover()` masks the panic, but the access is unsynchronized. Most likely to surface when an application registers a `NotifyClose` channel it does not continuously drain and the connection goes through a recoverable-then-terminal shutdown.

### Notes

Found while reviewing the topology-recovery work, but pre-existing on `main`: the `go func` send + `recover` path and the `close(listener)` are both present there as of 34bdaf0. Filing separately so it can be fixed on its own.

PR #361, opened 2026-06-30
Fix Args binding identity, recovery cleanup panic, and test HTTP leak

lukebakken:
## Summary

Three fixes against the topology-recovery work, intended to merge into `feature/implement-topology-recovery`. Each is a self-contained commit so they can be taken independently.

### Include `Args` in tracked binding removal and merge identity

`recordBinding` deduplicates by the full `(Queue, Exchange, Key, Args)` tuple, but `removeBinding` and `removeExchangeBinding` matched only on `(Queue/Source, Exchange/Destination, Key)`. Two bindings that share those fields but differ only in arguments are distinct and simultaneously valid (a headers exchange routes entirely on arguments with an empty routing key). Unbinding one purged every binding sharing the tuple, so the survivors, which still exist on the broker, were silently dropped from the tracking store and not recovered after a reconnect. The connection-wide merge in `getTopologyConfiguration` had the same defect.

Adds `reflect.DeepEqual` on `Args` to the removal predicates and the merge dedup, matching `recordBinding`. Covered by new unit tests (`TestRemoveBindingPreservesArgsDistinctBindings`, `TestGetTopologyConfigurationGlobalMergePreservesArgsDistinctBindings`), verified to fail without the fix.

### Guard connection cleanup against double-closing listeners

When connection recovery is exhausted, `OnConnectionClose` calls `cleanup()`, which closed the `closes` and `blocks` notify listeners unconditionally. On the path where recovery fails because topology cannot be recovered, `Reconnect` closes the transport mid-retry; the reader then raises a non-recoverable shutdown that already closes those listeners and sets `noNotify=true`. The subsequent `cleanup()` closed them a second time and panicked with "close of closed channel", crashing the process.

Gates the listener closing in `cleanup()` on `!c.noNotify`, the same sentinel `shutdown()` sets when it finalizes the listeners. Covered by a new integration test (`TestConnectionRecoveryExhaustionDoesNotPanic`) that fails topology recovery on every retry and asserts the connection reaches StateClosed without panicking; verified to panic without the fix.

### Disable keep-alives in integration HTTP test helper

The management API helper `baseCall` used a zero-value `http.Client`, relying on the shared default transport that pools idle keep-alive connections. An idle connection left over from the last admin call in a test outlives the test, and `goleak.VerifyTestMain` flags its `net/http` goroutines at suite teardown. Gives the client a dedicated transport with `DisableKeepAlives` set.

## Verification

- `go build`, `go vet`, `gofmt`: clean.
- Unit tests pass.
- Full recovery integration suite plus the new tests, race-enabled: 15/15 pass, goleak clean.
- Entire integration suite (`./...`): pass, no regressions.

## Related

While reviewing this work, a pre-existing data race in `Connection.shutdown` (present on `main`) was found and filed separately as #360.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -1862,7 +1862,7 @@ func (c *Connection) removeBinding(bc BindingConfig) {
 		oldBindings := config.Bindings
 		active := config.Bindings[:0]
 		for _, b := range oldBindings {
-			if b.Queue != bc.Queue || b.Key != bc.Key || b.Exchange != bc.Exchange {
+			if b.Queue != bc.Queue || b.Key != bc.Key || b.Exchange != bc.Exchange || !reflect.DeepEqual(b.Args, bc.Args) {
 				active = append(active, b)
 			}
 		}
@@ -1900,7 +1900,7 @@ func (c *Connection) removeExchangeBinding(ebc ExchangeBindingConfig) {
 		oldExchangeBindings := config.ExchangeBindings
 		active := config.ExchangeBindings[:0]
 		for _, eb := range oldExchangeBindings {
-			if eb.Destination != ebc.Destination || eb.Key != ebc.Key || eb.Source != ebc.Source {
+			if eb.Destination != ebc.Destination || eb.Key != ebc.Key || eb.Source != ebc.Source || !reflect.DeepEqual(eb.Args, ebc.Args) {
 				active = append(active, eb)
 			}
 		}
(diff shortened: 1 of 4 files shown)


connection.go
```diff
@@ -1419,12 +1419,21 @@ func (c *Connection) cleanup() {
...
-		close(listener)
-	}
+	// shutdown() closes the close/block listeners and sets noNotify=true on the
+	// non-recoverable path. When recovery is exhausted, a non-recoverable
+	// shutdown may already have run (e.g. the transport was closed mid-retry),
+	// so closing these again would panic with "close of closed channel". Only
+	// close listeners that shutdown() did not already finalize. In the
+	// pure-dial-failure case shutdown leaves them open and noNotify false, so
+	// cleanup() remains responsible for closing them here.
+	if !c.noNotify {
```
suchitd: Gating only `cleanup()` with `!c.noNotify` is not sufficient. The panic can also happen inside `shutdown()` itself on the second retry — before `cleanup()` is even called.

I'm fixing this now, I'll add more details soon

suchitd: | Scenario | `!c.noNotify` in `cleanup()` | `nil` after each close |
| :--- | :--- | :--- |
| **`shutdown()` → retry `shutdown()`** | ❌ `shutdown()` never checks `noNotify` — iterates the same already-closed slice | ✅ retry 2 finds `c.closes == nil`, iterates nothing |
| **`shutdown()` → `cleanup()`** | ✅ `noNotify=true` set by `shutdown()` → `cleanup()` skips | ✅ `cleanup()` finds `c.closes == nil`, iterates nothing |
| **`cleanup()` → `shutdown()` (race)** | ❌ `noNotify` not yet set when `cleanup()` runs → `shutdown()` iterates the already-closed slice | ✅ `cleanup()` nils the slice → `shutdown()` iterates nothing |

---

Setting `c.closes, c.blocks = nil, nil` immediately after the close loop is completely self-contained per invocation:

* **First Mover Wins:** Whoever runs first closes the channels and destroys the reference.
* **Second Mover No-Ops:** Whoever runs second finds `nil` and safely skips execution.

This approach eliminates the need for shared state flags, avoids brittle ordering assumptions, and drops any borrowed semantics. 

> ℹ️ **Go Optimization Note:**
> Iterating a `nil` slice in Go is explicitly safe and costs nothing, making this a highly idiomatic and robust fix for the race condition.

suchitd:
Thanks for identifying and fixing the issues @lukebakken !

PR #363, opened 2026-07-03
fix: forget auto-delete topology on last consumer/binding removal

### LINKED ISSUE #368: Evict auto-delete queues and exchanges from topology store to prevent stale resurrection during recovery
### Summary
When `TopologyRecovery` is enabled, the Go client maintains an in-memory store of declared queues, exchanges, and bindings so it can automatically recreate them if the connection drops. 

However, RabbitMQ handles **auto-delete** entities dynamically on the server side:
* An **auto-delete queue** is automatically deleted by the broker as soon as its last active consumer cancels or disconnects.
* An **auto-delete exchange** is automatically deleted by the broker as soon as its last binding (to a queue or another exchange) is removed.

The current bug is that the client-side topology recovery store is completely unaware of these broker-side lifecycle events. When an auto-delete entity is naturally cleaned up by the broker, the client continues to hold onto it in memory. If a network disruption occurs later, the client walks its topology store during recovery and **incorrectly resurrects** these explicitly deleted queues or exchanges back on the broker.

This behavior compounds aggressively when dealing with exchange-to-exchange bindings or cascading topologies, where a single cancellation should dissolve an entire chain (`Outer Exchange` ➔ `Inner Exchange` ➔ `Queue`), but instead results in the entire dead pipeline being recreated upon connection recovery.

### Current behavior
When an auto-delete queue or exchange meets its deletion condition (e.g., the last active consumer cancels), the broker deletes the object immediately, but the client retains it locally as a stale entry. 

**Result:** Upon connection recovery, a queue or exchange that was intentionally killed by architectural design is now permanently resurrected on the RabbitMQ broker, leaking resources until the application completely restarts.

### To Reproduce
1. Connect to the broker with topology recovery enabled.
2. Declare an auto-delete queue (e.g., `task.temporary.worker`).
3. Attach a consumer to the queue, process a payload, and then explicitly cancel the consumer subscription.
4. Verify on the RabbitMQ Management dashboard/CLI that the broker has naturally deleted the queue.
5. Simulate a network disruption (e.g., restart the RabbitMQ node or disconnect the network interface) to trigger client-side topology recovery.

### Expected behavior
The client-side topology store should actively evict auto-delete queues, exchanges, and downstream bindings as soon as the conditions that trigger their broker-side deletion are met. Upon network recovery, the client should not attempt to redeclare entities that were naturally torn down.

suchitd:
## Proposed Changes

Auto-delete queues and exchanges that have lost all their consumers or bindings are deleted by the broker immediately; tracking them further produces a stale entry that resurrects them incorrectly on recovery.

When topology recovery is enabled, auto-delete entities must be removed from the recovery store as soon as the condition that will trigger their broker-side deletion is met.

This commit implements that logic and also adds integration tests to cover:
- auto-delete queue forgotten only after the last consumer cancels (not on the first of two cancels)
- auto-delete exchange cascade-forgotten after QueueUnbind
- QueueDelete cascading to the source exchange
- ExchangeUnbind cascading to the source exchange
- full three-level chain (outerExchange→innerExchange→queue→consumer) where a single cancel cascades all three entities, verified both in the topology store and on the broker after connection recovery

## Types of Changes

- [ ] Bugfix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

## Further Comments

This is a follow-up PR of https://github.com/rabbitmq/amqp091-go/pull/357

https://github.com/rabbitmq/amqp091-go/issues/368

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -1760,11 +1760,13 @@ func (c *Connection) recordExchange(channelID uint16, ec ExchangeConfig) {
 	c.channelTopology(channelID).Exchanges[ec.Name] = ec
 }
 
-// removeExchangeLocked removes an exchange and its bindings from the topology store.
+// removeExchange removes an exchange and its bindings from the topology store.
 // It returns the exchange names that were sources in exchange-to-exchange bindings
 // pointing TO the deleted exchange — these are candidates for auto-delete cascade.
-// Caller must hold c.topologyM.
-func (c *Connection) removeExchangeLocked(name string) []string {
+func (c *Connection) removeExchange(name string) []string {
+	c.topologyM.Lock()
+	defer c.topologyM.Unlock()
+
 	if c.topologyConfiguration == nil {
 		return nil
 	}
@@ -1806,19 +1808,10 @@ func (c *Connection) removeExchangeLocked(name string) []string {
 	return cascadeSources
 }
 
-func (c *Connection) removeExchange(name string) {
-	c.topologyM.Lock()
-	defer c.topologyM.Unlock()
-	c.removeExchangeLocked(name)
-}
-
 // deleteRecordedExchange removes an exchange from tracking and cascades to any
 // auto-delete exchange that sourced an exchange-to-exchange binding pointing to it.
 func (c *Connection) deleteRecordedExchange(name string) {
-	c.topologyM.Lock()
-	cascadeSources := c.removeExchangeLocked(name)
-	c.topologyM.Unlock()
-	for _, src := range cascadeSources {
+	for _, src := range c.removeExchange(name) {
 		c.maybeDeleteRecordedAutoDeleteExchange(src)
 	}
 }
@@ -1872,10 +1865,13 @@ func (c *Connection) recordQueue(channelID uint16, qc QueueConfig) {
 	c.channelTopology(channelID).Queues[qc.ActualName] = qc
 }
 
-// removeQueueLocked removes a queue and its bindings from the topology store.
+// removeQueue removes a queue and its bindings from the topology store.
 // It returns the exchange names that sourced bindings pointing to this queue —
-// these are candidates for auto-delete cascade. Caller must hold c.topologyM.
-func (c *Connection) removeQueueLocked(name string) []string {
+// these are candidates for auto-delete cascade.
+func (c *Connection) removeQueue(name string) []string {
+	c.topologyM.Lock()
+	defer c.topologyM.Unlock()
+
 	if c.topologyConfiguration == nil {
 		return nil
 	}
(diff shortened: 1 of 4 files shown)


connection.go
```diff
@@ -1833,6 +1896,56 @@ func (c *Connection) removeQueue(name string) {
...
+	for _, ch := range c.channels {
+		channels = append(channels, ch)
+	}
+	c.m.Unlock()
+
+	for _, ch := range channels {
+		if ch.consumers.hasConsumerForQueue(queueName) {
+			return
+		}
+	}
```
Zerpet, line 1945: I'm not liking this too much. This call locks `consumers`, in addition to creating a copy of the channel slices (not too bad since they are pointers) earlier.

I prefer the approach taken in the .NET client, where it simply checks its own topology recovery records, and tries to determine if it has a consumer record to queue X.

It is the same approach as in here, but using exclusively the recovery records, which I think is the right thing to do. You also avoid locking `consumers`, which is already shared by `Channel`.

Zerpet: Dotnet client reference:

https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/c702061cf648f77851092d194bbd95fda17339a8/projects/RabbitMQ.Client/Impl/AutorecoveringConnection.Recording.cs#L403-L435

suchitd: In this client, the consumers are tracked at [channel level](https://github.com/rabbitmq/amqp091-go/blob/main/channel.go#L73) from the start. For topology recovery records, I chose not make any changes there as they(consumers) are genuinely channel scoped.  Thats the reason you see what you see :) 

I'ven't gone through .NET client, but based on your comment, if we want to achieve that, the changes would be to move consumers records at connection level, which I think certainly out of scope of this PR. Please let me know.

> I'm not liking this too much. This call locks consumers

Why is this an issue? Consumers are initialised [per channel level](https://github.com/rabbitmq/amqp091-go/blob/main/channel.go#L223), and it locks for that particular instance of channel. Also, `hasConsumerForQueue` is simply iterating over the map and returning the response, hence the lock will be held for a negligible time. All other consumer methods that contends for consumer lock are quick in response(no heavy processing) and I don't see a place where consumer lock contention may occur.

Zerpet: I didn't express clearly, my bad. I don't dislike the mutex call, but the fact that we are consulting the consumer state (consumerConfig), instead of the recovery records.

I understand that this client can't copy 1-for-1 the .NET client. I re-read the code and I see the difficulty now: each channel is only aware of its consumers, and different channels can have consumers to the same AD queue.

I agree that refactoring the topology recovery records is out of the scope for this PR. We should note the current logic is coupling channel with connection (even more). Note as well that, for each channel, it consults the connection to get a list of all channels, to determine if there are consumers for a certain queue. In other words, _each channel consults the state of all channels_.

Zerpet (approved):
Let's go ahead and merge, we can refactor later.

connection.go
suchitd: Yes, thats very true! I've made a note of it, I'll fix this through a follow-up PR.

PR #365, opened 2026-07-09
feat: skip-and-continue topology recovery with per-entity error surfacing

suchitd:
## Proposed Changes

Topology recovery previously aborted on the first entity failure, causing
the entire retry cycle to restart even when a single exchange or queue
was missing or mis-configured. This change introduces a skip-and-continue
model with fine-grained error visibility.

**Structural Removal of `isRecoverable`:**
To implement a resilient "Skip and Continue" topology recovery architecture, 
channels must not be completely torn down on soft broker exceptions 
(such as a 406 PRECONDITION_FAILED due to argument mismatches). Maintaining 
a hardcoded list of recoverable error codes is incredibly fragile, complex, 
and prone to hidden regressions if a specific broker or network error frame 
is missed.

To establish a clean separation of concerns, the `isRecoverable(err)` error 
classification layer has been completely removed from the connection and 
channel shutdown phases. 

Instead of forcing a single method to handle mixed-mode teardowns based on 
volatile error codes, responsibilities are cleanly bifurcated based on 
whether recovery is globally enabled:

1. **When Recovery is Enabled**: Unexpected connection-level drops (`err.Code >= 500`, 
   `320`, or `FrameError`) trigger minimal transport-only closures (`close(ch.errors)`, 
   `close(ch.close)`). This safely unblocks any blocking network RPC threads while 
   preserving the client-side consumer layouts and channel registry maps (`c.channels`) 
   perfectly in memory.
2. **When Recovery is Not Enabled / Graceful Close**: User-initiated closures 
   (where `err == nil`) or localized channel exceptions (`406` / `405`) bypass the 
   recovery shunts and immediately execute full, destructive teardowns.
3. **Recovery Lifecycle Ownership**: If automatic recovery fails permanently 
   and exhausts all retries, the recovery framework assumes total responsibility 
   for terminal destruction by invoking `Connection.cleanup()`. It leverages the 
   safely preserved maps to forcefully wipe remaining consumers, close notification 
   pipes, broadcast the final terminal `StateClosed` state, and eliminate any 
   lingering `goleak` goroutine failures.

## Types of Changes

- [ ] Bugfix 
- [x] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

## Further Comments

Connection and Topology recovery feature is in Experimental phase and changes are expected to make it more resilient.

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -253,6 +253,8 @@ func (ch *Channel) shutdown(e *Error) {
 	ch.notifyM.Lock()
 	defer ch.notifyM.Unlock()
 
+	isRecoveryInProgress := ch.connection.inTopologyRecovery.Load()
+	Logger.Printf("Channel isRecoveryInProgress: %v", isRecoveryInProgress)
 	// Broadcast abnormal shutdown
 	if e != nil {
 		for _, c := range ch.closes {
@@ -281,7 +283,8 @@ func (ch *Channel) shutdown(e *Error) {
 		}
 	}
 
-	if e == nil || !ch.connection.IsRecoveryEnabled() || !ch.connection.isRecoverable(e) {
+	if e == nil || !ch.connection.IsRecoveryEnabled() || (!ch.connection.isRecoverable(e) && !isRecoveryInProgress) {
+		Logger.Printf("Channel: Removing consumers")
 		ch.consumers.close()
 
 		for _, c := range ch.closes {
@@ -2203,6 +2206,8 @@ func (ch *Channel) cleanup() {
 	ch.notifyM.Lock()
 	defer ch.notifyM.Unlock()
 
+	Logger.Printf("Cleanup channel")
+
 	ch.consumers.close()
 
 	for _, c := range ch.closes {
@@ -2276,6 +2281,31 @@ func (ch *Channel) Reconnect() error {
 	return nil
 }
 
+// openChannelSession resets client-side state, opens a fresh broker channel,
+// and restores QoS/Confirm configuration. It is the shared single-attempt core
+// used by both reconnectChannel (retry loop) and reopenChannelIfClosed (topology
+// recovery). The caller must hold ch.reconnecting and manage lifecycle transitions.
+//
+// openSucceeded is true when ch.open() completed — the caller needs this to decide
+// whether to send a channel.close courtesy frame to the broker before the next
+// retry (meaningful only when open succeeded but setup then failed).
+func (ch *Channel) openChannelSession() (openSucceeded bool, err error) {
+	// 1. Reset client-side state
+	ch.destructorM.Lock()
+	ch.m.Lock()
+	ch.resetState()
+	ch.m.Unlock()
+	ch.destructorM.Unlock()
+
+	// 2. Open a fresh channel on the broker
+	if err = ch.open(); err != nil {
+		return false, err
+	}
+
+	// 3. Perform QoS and Confirms setup.
+	return true, ch.setupChannelBasic()
+}
+
 // reconnectChannel opens a fresh channel on the broker and performs basic setup (QoS, Confirms).
 // It does NOT recover the channel's topology.
 func (ch *Channel) reconnectChannel() error {
@@ -2294,7 +2324,10 @@ func (ch *Channel) reconnectChannel() error {
 
 	cancelCh := ch.NotifyRecoveryCancel(make(chan struct{}))
 
-	var err error
+	var (
+		err    error
+		opened bool
+	)
 	for i := 0; i < ch.connection.MaxRetryCount(); i++ {
 		// Exit early if Close() was already called
 		select {
@@ -2317,24 +2350,16 @@ func (ch *Channel) reconnectChannel() error {
 			}
 		}
 
-		// 1. Reset client-side state
-		ch.destructorM.Lock()
-		ch.m.Lock()
-		ch.resetState()
-		ch.m.Unlock()
-		ch.destructorM.Unlock()
-
-		// 2. Open a fresh channel on the broker
-		if err = ch.open(); err != nil {
-			Logger.Printf("Channel %d recovery open error: %v", ch.id, err)
-			continue
-		}
-
-		// 3. Perform QoS and Confirms setup.
-		if err = ch.setupChannelBasic(); err != nil {
-			Logger.Printf("Channel %d setup failed: %v, closing and retrying...", ch.id, err)
-			ch.setClosed()
-			_ = ch.call(&channelClose{ReplyCode: replySuccess, ReplyText: "Recovery retry"}, &channelCloseOk{})
+		opened, err = ch.openChannelSession()
+		if err != nil {
+			Logger.Printf("Channel %d recovery attempt %d failed: %v", ch.id, i+1, err)
+			if opened {
+				// open() succeeded but setupChannelBasic() failed:
+				// gracefully close the broker-side session before the next attempt.
+				// channelClose must be sent before setClosed()
+				_ = ch.call(&channelClose{ReplyCode: replySuccess, ReplyText: "Recovery retry"}, &channelCloseOk{})
+				ch.setClosed()
+			}
 			continue
 		}
 
--- a/channel.go
+++ b/channel.go
@@ -114,6 +114,14 @@ type Channel struct {
 	reconnecting sync.Mutex // Mutex for reconnecting channel.
 	lifeCycle    *lifeCycle // The current state of the channel.
 
+	// recoveringTopology is true while a recoverConnectionTopology pass owns
+	// this channel's reopen/redeclare sequence. watchChannel's NotifyClose
+	// listener checks this to avoid starting a redundant, competing
+	// Reconnect+RecoverTopology pass when a broker soft error (e.g. a
+	// PRECONDITION_FAILED from an entity the in-flight pass is already
+	// handling) closes the channel out from under it.
+	recoveringTopology atomic.Bool
+
 	recoveryCancels []chan struct{} // listeners for channel recovery cancellation
 }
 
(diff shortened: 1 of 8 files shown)


channel.go
```diff
@@ -2236,6 +2248,13 @@ func (ch *Channel) cleanup() {
 	}
 
 	ch.noNotify = true
+
+	var err error
+	if e != nil {
+		err = fmt.Errorf("%w", e)
```
Zerpet:
```suggestion
		err = fmt.Errorf("error in topology cleanup: %w", e)
```

Wrapping an error without providing additional context is an anti-pattern. If there's no additional context to provide, it's better to return as-is.

connection.go
```diff
@@ -1473,6 +1476,13 @@ func (c *Connection) cleanup() {
 	c.topologyM.Lock()
 	c.topologyConfiguration = nil
 	c.topologyM.Unlock()
+
+	var e error
+	if err != nil {
+		e = fmt.Errorf("%w", err)
```
Zerpet:
```suggestion
		e = fmt.Errorf("error in connection cleanup: %w", err)
```

Same as above, we should add context when wrapping an error.

connection.go
```diff
@@ -2277,16 +2233,79 @@ func (c *Connection) recoverConnectionTopology(channels []*Channel) error {
...
+	// skipOrAbort calls cb with the connection and entity error. If cb returns true
+	// (or is nil, matching DefaultOnTopologyEntityError behaviour) the entity is
+	// appended to skipped and the caller continues the loop. If cb returns false
+	// the entity error is returned as a fatal error, aborting topology recovery.
+	skipOrAbort := func(e TopologyRecoveryEntity) (skip bool, fatal error) {
+		if cb == nil || cb(c, e) {
+			skipped = append(skipped, e)
+			return true, nil
+		}
+		return false, fmt.Errorf("%w", e)
```
Zerpet, line 2268: Same as above. Either add context information or return the error as-is.

connection.go
```diff
@@ -2277,16 +2233,79 @@ func (c *Connection) recoverConnectionTopology(channels []*Channel) error {
...
+	}
+
+	// reopenChannelIfClosed reopens a channel that was closed as a side-effect of a
+	// broker soft error (e.g. PRECONDITION_FAILED / 406) during topology recovery.
+	reopenChannelIfClosed := func(ch *Channel) {
+		// Fast path: skip the mutex entirely for the common case where the channel
+		// is already open. Avoids blocking on ch.reconnecting for every entity in
+		// a healthy recovery where no reopen is needed.
```
Zerpet: I wonder why this can be a `Channel` function, instead of an anonymous function?

For example `func (Channel* ch) reopenChannelIfClosed(...) {}`

Zerpet:
> **Note:** This review was generated by an AI coding agent (Claude Code) with a human reviewer in the loop. The human reviewed and approved the findings below before posting; the underlying claims — including the reconnect-loop repro — were independently verified against a live RabbitMQ instance rather than taken at face value.

## Review: skip-and-continue topology recovery

Thanks for tackling this — the skip-and-continue direction is a real improvement over
aborting recovery on the first entity failure. Below is a review of the diff, with the
riskier changes verified against a live RabbitMQ (Docker) rather than just read.
Worst issue first.

### 🔴 Confirmed: skipping a permanently-failing entity causes an infinite reconnect loop

Repro: dialed with `Recovery` enabled and `OnTopologyEntityError` returning `true`
(skip), declared a channel with two durable queues, then from a second connection
redeclared both queues out-of-band with a conflicting `x-max-length` (a permanent,
realistic conflict — e.g. topology changed by another process). Forced one soft error
on the channel to kick off recovery, then watched `RecoverTopology` invocation counts
over 30s.

Instead of settling into a stable open channel with two skipped entities, the channel
enters a permanent cycle, roughly every 5 seconds, forever:

```
Initiating channel 1 recovery → RecoverTopology call #N → skip q1, skip q2 → channel reopened →
(5s later) Channel 1 closed unexpectedly [q2's *already-skipped* error] → Initiating channel 1 recovery → call #N+1 → ...
```

I instrumented a wrapper around `TopologyRecovery.RecoverTopology` to count concurrent
calls (I originally suspected a race between `watchChannel`'s listener and the
explicit `reopenChannelIfClosed` reopen — that part is *not* the problem: max
concurrent calls was always 1, and `-race` reported nothing). The real cause is more
subtle: the 5-second period is exactly `notifyTimeout` (`channel.go:28`, used at
`channel.go:273`). `Channel.shutdown` always broadcasts the closing error onto
`ch.closes` — including via the "buffer full, deliver in a background goroutine with a
5s timeout" fallback — *regardless* of whether the teardown branch below it runs.
Since that teardown branch is skipped whenever recovery is enabled and `e != nil`,
`watchChannel`'s `NotifyClose` listener is never torn down and stays live across the
entire recovery pass. It ends up receiving the *same, already-skipped* entity error
again via that 5s fallback, and re-triggers `OnChannelClose` → `ch.Reconnect()` → a
brand new `RecoverTopology` pass that rediscovers and re-skips the same entities —
indefinitely.

This undermines the point of skip-and-continue: for entities that are *actually*
permanently broken (the realistic case for using `OnTopologyEntityError`), the channel
never reaches a stable state — it hammers the broker with redundant redeclares every
~5s with no backoff and no way to stop short of closing the channel/connection. I'd
consider this blocking.

Possible fix direction: once a `RecoverTopology` pass completes (skipped or not) and
the channel is left open, the notification that triggered this pass needs to be
drained/consumed before `watchChannel` is allowed to react to it again — or
`Channel.Reconnect`/`reconnectChannel` needs to recognize "I'm already handling this
exact close" and no-op instead of starting a fresh pass.

The repro program and full log are posted in a follow-up comment.

### 🟠 Channel registry/ID leak when a single channel's recovery exhausts retries

`closeChannel` (connection.go) only calls `releaseChannel` when
`e == nil || !c.IsRecoveryEnabled()`. When recovery is enabled and a channel closes
with an error, it's deliberately kept in `c.channels` (per the comment, so
`reopenChannelIfClosed`/`connection.cleanup` can find it later). But when *per-channel*
recovery exhausts `MaxRetryCount`, `DefaultConnectionRecovery.OnChannelClose` calls
`ch.cleanup(amqpErr)` directly — and `Channel.cleanup` never touches `c.channels` or
`c.allocator`. A channel that permanently fails to recover is left as a dead entry in
`Connection.channels` forever, and its ID is never returned to the allocator. On a
long-lived connection with recovery enabled, repeated permanent per-channel failures
will leak channel IDs.

Suggest: have `Channel.cleanup` (or its caller) also release the channel from the
connection's registry/allocator when the whole connection isn't already being torn
down.

### 🟠 Removing `isRecoverable`/`RecoverableErrorCodes` widens automatic recovery to *all* channel errors — and it's a breaking API change

With `isRecoverable` gone, `watchChannel`'s listener now fires `OnChannelClose` →
`ch.Reconnect()` for *any* channel-level error once `Recovery` is configured —
including ordinary soft protocol errors like `406 PRECONDITION_FAILED` or
`405 RESOURCE_LOCKED` from a mismatched declare, which previously were correctly left
alone (this was exactly what the now-deleted `TestConnectionRecoveryNonRecoverableChannelClose`
covered). The app still gets its error back from the call, but the channel it's
holding a reference to now also gets silently torn down and reopened behind its back.
That's a significant, easy-to-miss behavior change and should be called out explicitly
in the PR description/changelog, not just implied by the code removal.

Separately: `Connection.GetRecoverableErrorCodes`, `SetRecoverableErrorCodes`,
`AddRecoverableErrorCodes`, `ReconnectionConfig.RecoverableErrorCodes`, and
`ErrRecoveryNotEnabled` are all present and exported in the latest release (`v1.12.0`)
— confirmed directly (`git show v1.12.0:connection.go` / `recovery.go`). Removing them
outright breaks source compatibility for anyone building against `v1.12.0`. Given the
feature is still experimental, an outright removal might be acceptable, but the more
conventional path would be to mark them `// Deprecated: ...` (pointing at
`OnTopologyEntityError`/the new skip-and-continue model as the replacement) for one
release cycle instead of deleting them in the same PR that changes their underlying
semantics. That also gives downstream users a compiler-visible signal before the
behavior shifts under them.

(Note: I initially also flagged the `TopologyRecovery` interface signature change —
`error` → `([]TopologyRecoveryEntity, error)` — as breaking. That one's fine as-is:
checked every tag back to `v1.8.1`, and `TopologyRecovery` doesn't exist in any
released version, so changing its shape freely here is not a compatibility concern.)

### Minor
- `fmt.Errorf("%w", e)` in `Channel.cleanup`/`Connection.cleanup` wraps an `*Error`
  (which already implements `error`) purely to change its static type — `err = e`
  would do, unless the wrapping is intentionally hiding the concrete type from
  `StateChanged.Err` consumers, in which case a short comment explaining why would
  help future readers (downstream code already does `errors.As` to unwrap it anyway).
- `TopologyRecoveryEntity.Error()` formats `%v` on `e.Err` without a nil-guard; not
  reachable from current call sites, but the type is public and constructible by
  callers.

### Test coverage
- `TestConnectionRecoverySkipAndContinue` and the updated
  `TestConnectionRecoveryExhaustionDoesNotPanic` are good additions and read well.
- Given the livelock above, suggest a test that declares a permanently-conflicting
  entity (as in the repro), lets recovery run for e.g. 10-15s, and asserts the number
  of recovery attempts is bounded (or that the channel settles into `StateOpen` and
  stays there) — this would have caught the loop.

Zerpet:
Repro program and full log for the reconnect-loop finding above: https://gist.github.com/Zerpet/51f3ee912ac860ed90732920c5772615

channel.go
suchitd: Done!

connection.go
suchitd: Done!

connection.go
suchitd: Done!

connection.go
suchitd: Done!

suchitd:
> Repro program and full log for the reconnect-loop finding above: https://gist.github.com/Zerpet/51f3ee912ac860ed90732920c5772615

Thanks for the review @Zerpet ! I've pushed the fix to this PR. 
I re-ran your example with the fix and it does not loop now. I've also written an integration test inspired by your example.

PR #366, opened 2026-07-10
fix: eliminate multiple data races in Channel and Connection operations

### LINKED ISSUE #360: Data race in Connection.shutdown between buffered listener send goroutine and close(listener)

suchitd:
## Proposed Changes

Channel:
- call() and dispatch() read ch.errors, ch.rpc, and ch.close without holding ch.m. Fixed by snapshotting the fields under ch.m before entering the select.
- ch.confirming was read from multiple goroutines without synchronization. Replaced the bool + confirmM mutex with atomic.Bool; all five read sites updated to .Load().
- shutdown() background goroutines (for full listener channels) could send to a listener while cleanup() concurrently closed it. Fixed with the existing notifyM RWMutex: shutdown() holds notifyM.Lock() while spawning goroutines; goroutines hold notifyM.RLock() during the send; cleanup() holds notifyM.Lock() before closing listener channels — send and close are mutually exclusive.
- shutdown() background goroutines (for full listener channels) could send to a listener while cleanup() concurrently closed it. Fixed with the existing notifyM RWMutex: shutdown() holds notifyM.Lock() while spawning goroutines; goroutines hold notifyM.RLock() during the send; cleanup() holds notifyM.Lock() before closing listener channels — send and close are mutually exclusive.

Connection:
- call() read c.errors and c.rpc without holding c.m. Fixed with the same field-snapshot pattern as Channel.call().
- dispatch0() read c.blocks without holding c.m before notifyAll. Fixed by snapshotting c.blocks under c.m.
- reader()'s deferred close(c.rpc) captured the field at return time, racing with resetState(). Fixed by capturing rpc := c.rpc at function entry.
- shutdown() background goroutines raced with cleanup() on c.closes for the same reason as Channel. Fixed by adding a dedicated notifyM RWMutex to Connection (mirroring Channel's pattern) with the same lock ordering: destructorM → m → notifyM.

## Types of Changes

- [x] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

## Fixes

- https://github.com/rabbitmq/amqp091-go/issues/360

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -255,7 +255,6 @@ func (ch *Channel) shutdown(e *Error) {
 
 	// Broadcast abnormal shutdown
 	if e != nil {
-		closeCh := ch.close // capture before shutdown closes it; used as abort signal
 		for _, c := range ch.closes {
 			select {
 			case c <- e:
@@ -264,19 +263,15 @@ func (ch *Channel) shutdown(e *Error) {
 				// the shutdown sequence. The goroutine holds notifyM.RLock() for the
 				// duration of the send, which is mutually exclusive with cleanup()'s
 				// notifyM.Lock(), preventing a concurrent send+close data race.
-				// shutdown() holds notifyM.Lock() right now, so the goroutine is blocked
-				// on RLock until shutdown() closes ch.close (the abort) and returns —
-				// guaranteeing <-done is always ready when the goroutine first runs.
-				go func(c chan *Error, e *Error, done <-chan struct{}) {
+				go func(c chan *Error, e *Error) {
 					defer func() { _ = recover() }()
 					ch.notifyM.RLock()
 					defer ch.notifyM.RUnlock()
 					select {
 					case c <- e:
-					case <-done:
 					case <-time.After(5 * time.Second):
 					}
-				}(c, e, closeCh)
+				}(c, e)
 			}
 		}
 		// Notify RPC if we're selecting
--- a/channel.go
+++ b/channel.go
@@ -64,7 +64,6 @@ type Channel struct {
 	destructorM sync.Mutex   // Mutex for destroying the channel.
 	destructed  bool         // Will be true if the channel has been destroyed, false otherwise.
 	m           sync.Mutex   // Mutex for the channel.
-	confirmM    sync.Mutex   // Mutex for the publisher confirms state.
 	notifyM     sync.RWMutex // Mutex for the notify state.
 
 	connection *Connection
@@ -98,7 +97,7 @@ type Channel struct {
 
 	// Allocated when in confirm mode in order to track publish counter and order confirms
 	confirms   *confirms
-	confirming bool
+	confirming atomic.Bool
 
 	// Selects on any errors from shutdown during RPC
 	errors chan *Error
@@ -256,21 +255,28 @@ func (ch *Channel) shutdown(e *Error) {
 
 	// Broadcast abnormal shutdown
 	if e != nil {
+		closeCh := ch.close // capture before shutdown closes it; used as abort signal
 		for _, c := range ch.closes {
 			select {
 			case c <- e:
 			default:
-				// If blocked/full, send in a goroutine so we never deadlock the shutdown sequence
-				go func(c chan *Error, e *Error) {
-					defer func() {
-						_ = recover() // Gracefully ignore panics if the channel is closed concurrently
-					}()
+				// Channel is full; deliver in a background goroutine so we never deadlock
+				// the shutdown sequence. The goroutine holds notifyM.RLock() for the
+				// duration of the send, which is mutually exclusive with cleanup()'s
+				// notifyM.Lock(), preventing a concurrent send+close data race.
+				// shutdown() holds notifyM.Lock() right now, so the goroutine is blocked
+				// on RLock until shutdown() closes ch.close (the abort) and returns —
+				// guaranteeing <-done is always ready when the goroutine first runs.
+				go func(c chan *Error, e *Error, done <-chan struct{}) {
+					defer func() { _ = recover() }()
+					ch.notifyM.RLock()
+					defer ch.notifyM.RUnlock()
 					select {
 					case c <- e:
+					case <-done:
 					case <-time.After(5 * time.Second):
-						// Give up to avoid leaking the goroutine permanently
 					}
-				}(c, e)
+				}(c, e, closeCh)
 			}
 		}
 		// Notify RPC if we're selecting
@@ -350,14 +356,19 @@ func (ch *Channel) call(req message, res ...message) error {
 	}
 
 	if req.wait() {
+		ch.m.Lock()
+		errors := ch.errors
+		rpc := ch.rpc
+		ch.m.Unlock()
+
 		select {
-		case e, ok := <-ch.errors:
+		case e, ok := <-errors:
 			if ok {
 				return e
 			}
 			return ErrClosed
 
-		case msg := <-ch.rpc:
+		case msg := <-rpc:
 			if msg != nil {
 				for _, try := range res {
 					if reflect.TypeOf(msg) == reflect.TypeOf(try) {
@@ -514,7 +525,7 @@ func (ch *Channel) dispatch(msg message) {
 		ch.notifyM.RUnlock()
 
 	case *basicAck:
-		if ch.confirming {
+		if ch.confirming.Load() {
 			if m.Multiple {
 				ch.confirms.Multiple(Confirmation{m.DeliveryTag, true})
 			} else {
@@ -523,7 +534,7 @@ func (ch *Channel) dispatch(msg message) {
 		}
 
 	case *basicNack:
-		if ch.confirming {
+		if ch.confirming.Load() {
 			if m.Multiple {
 				ch.confirms.Multiple(Confirmation{m.DeliveryTag, false})
 			} else {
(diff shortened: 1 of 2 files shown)


MirahImage (approved):

Zerpet:
**Claude review below** -- the first observation is correct, regarding the inefficiency of the background goroutine. The second is cosmetic, as Claude suggests, but I also caught it during my review (without AI).

----

## Review

### Overview
Addresses issue #360 by fixing several data races flagged by `-race` in `Channel`/`Connection`:
- Snapshotting `errors`/`rpc` (and `blocks`) fields under the struct mutex before `select`-ing on them in `call()`, `dispatch()`, and `dispatch0()`.
- Converting `Channel.confirming` from `bool` + dedicated mutex to `atomic.Bool`.
- Adding a `notifyM sync.RWMutex` to `Connection` (mirroring the existing `Channel.notifyM`) to serialize `shutdown()`'s background delivery goroutines against `cleanup()`'s channel closes.

Most of this is solid and directly addresses real races. One change, however, looks like it silently defeats an existing safety mechanism.

### Correctness issues

**1. The new `done` abort channel in the background delivery goroutines likely nullifies the 5-second delivery window (High)**

In both `Channel.shutdown()` and `Connection.shutdown()`, the "listener channel is full" fallback goroutine now does:

```go
go func(c chan *Error, e *Error, done <-chan struct{}) {
    defer func() { _ = recover() }()
    ch.notifyM.RLock()
    defer ch.notifyM.RUnlock()
    select {
    case c <- e:
    case <-done:
    case <-time.After(5 * time.Second):
    }
}(c, e, closeCh)
```

`done` is `ch.close` (or `c.errors` for `Connection`). But `ch.close`/`c.errors` is unconditionally closed *inside the same `shutdown()` call*, before `shutdown()` returns and releases `notifyM.Lock()`. Since the goroutine can't acquire `notifyM.RLock()` until `shutdown()` releases the write lock, `done` is **already closed** the very first time the goroutine's `select` runs. Unless `c <- e` happens to also be simultaneously ready (i.e., the consumer is actively draining at that exact instant), Go's `select` will just take the `<-done` case immediately — the goroutine exits without ever giving the slow/full listener the originally-intended up-to-5-seconds chance to drain and receive the final shutdown error.

Effectively, this turns "retry delivery for 5s" into "give up almost immediately" for every abnormal shutdown with a full listener channel. The mutual exclusion from `notifyM` (RLock in the goroutine vs. Lock in `cleanup()`) is what actually prevents the send-after-close panic — that path was already guarded by the pre-existing `recover()` regardless. The `done` case doesn't add safety; it just guts the retry window.

The race described in the PR body ("shutdown() background goroutines … could send to a listener while cleanup() concurrently closed it") is real, and the `notifyM` locking legitimately fixes it. But `done` should probably be tied to something that's only closed by the *other* code path (`cleanup()`), not something `shutdown()` itself closes before releasing the lock the goroutine waits on.

**2. `reader()`'s "fix" is a no-op (Low, but the stated rationale is incorrect)**

```go
-	defer close(c.rpc)
+	rpc := c.rpc
+	defer close(rpc)
```

The description says this fixes `c.rpc` being "captured at return time, racing with `resetState()`." That's not how Go's `defer` works — `defer close(c.rpc)` already evaluates `c.rpc` immediately when the `defer` statement executes (at function entry here), not at return time. The new code evaluates `c.rpc` at the same point, just via an explicit local, and still does so without holding `c.m`. Functionally identical to before — harmless, but the write-up overstates what changed, and if there really is a race here it isn't fixed by this diff.

### Other observations

- **`atomic.Bool` conversion for `confirming`**: clean, all five read/write sites converted consistently, no leftover references to the removed `confirmM`. Good mechanical fix.
- **`ch.errors`/`ch.rpc`/`c.errors`/`c.rpc`/`c.blocks` snapshotting under the struct mutex**: correct and consistent with the documented invariant that `resetState()` requires the mutex held.
- **PR description vs. diff mismatch**: the description mentions a `cleanedUp.CompareAndSwap` gate used to prevent `shutdown()`/`cleanup()` double-teardown, but no such field exists in this diff or the current codebase — the actual mechanism used is the `notifyM` RWMutex plus the pre-existing `destructed` flag. Worth updating the description so reviewers aren't misled about the mechanism actually implemented.
- **No new tests**: a PR fixing five distinct data races (verified only by "tests pass locally") ships with no `-race` regression test that would fail without the fix and pass with it. At least one targeted test (e.g., concurrent `Confirm()` + publish, or concurrent shutdown + slow listener) would materially increase confidence and guard against regressions.
- **Lock ordering**: `m` → `notifyM` ordering is applied consistently in both `shutdown()`/`cleanup()` pairs for `Channel` and `Connection` — no new inversion introduced.

### Suggestions
1. Reconsider the `done` channel in the delivery goroutines — either drop it (relying solely on `notifyM` for the send/close race) or tie it to a signal actually controlled by the concurrent closer (`cleanup()`), not one closed by the same `shutdown()` call that spawned the goroutine.
2. Correct or drop the `reader()` change, since the stated race doesn't exist under Go's defer-argument-evaluation semantics.
3. Add at least one `-race`-tagged regression test exercising the confirm-mode and full-listener-channel shutdown scenarios this PR targets.
4. Update the PR description to reflect the actual `notifyM`-based mechanism rather than the `cleanedUp.CompareAndSwap` approach it currently describes.

channel.go
```diff
@@ -355,14 +361,19 @@ func (ch *Channel) call(req message, res ...message) error {
 	}
 
 	if req.wait() {
+		ch.m.Lock()
+		errors := ch.errors
+		rpc := ch.rpc
```
Zerpet, line 361: I am unclear how "snapshotting" prevents the race?
Channels and slices are passed by reference, therefore, accessing `errors` or `ch.errors` is the same, it does not address the actual race. The right thing to do would be to block/wait closing the channels until we have delivered the notifications or given up, whichever happens first.

channel.go
```diff
@@ -542,10 +553,15 @@ func (ch *Channel) dispatch(msg message) {
 		// deliveries are in flight and a no-wait cancel has happened
 
 	default:
+		ch.m.Lock()
+		closeCh := ch.close
+		rpc := ch.rpc
+		ch.m.Unlock()
```
Zerpet, line 558: Same as above.

Zerpet (changes requested):
Human review below 👨‍💻

channel.go
suchitd: The race is on the pointer/reference itself and not on the content.
`ch.errors` and `ch.rpc` are pointer fields pointing to underlying channels. These references can be updated during [resetState](https://github.com/rabbitmq/amqp091-go/blob/main/channel.go#L2356-L2358) concurrently when recovery is in progress and some other goroutine tries to invoke Channel.* that internally calls `call()`.

channel.go
suchitd: Please refer to my above comment. Thanks!

channel.go
Zerpet: > The race is on the pointer/reference itself and not on the content.

I'm not sure it makes a difference, in either case, access to the field, either to write the field or to "use" the field, must be protected. I don't think we should focus on that. The million dollar question is: why other goroutines try to call `Channel.call()` when recovery is in progress?

And the answer is `Channel.Reconnect()`. Since it is exported, it can be called from the app, and at the same time, other channel operations e.g. `Channel.QueueDeclare()` can be called from the app. That's where the race for channel fields is coming from. The "snapshotting" ensures that `call()` uses the fields it originally intended to use, even if `resetState()` writes the fields before `call()` returns.

I understand now 😄

Zerpet (approved):
Please squash or rebase on top of latest `main` before merging.

PR #369, opened 2026-07-14
fix: reject frames exceeding negotiated frame_max before allocation

suchitd:
## Proposed Changes

A malicious or compromised broker could declare an oversized length in a frame header (e.g. within a basic.deliver content body frame) and force the client to allocate memory for it, bypassing the frame_max negotiated during connection.tune and enabling a memory-exhaustion DoS.

Track the negotiated frame_max in a new atomic Connection.maxFrameSize field, mirrored from Config.FrameSize once connection.tune completes, and check every incoming frame's declared size against it in reader.ReadFrame before dispatching to the type-specific parsers that allocate the payload buffer. Frames over the limit are rejected with ErrFrameTooLarge instead of being read.

## Types of Changes

- [x] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

- [ ] I have added necessary documentation (if appropriate)

### DIFF
--- a/read.go
+++ b/read.go
@@ -54,6 +54,12 @@ func (r *reader) ReadFrame() (frame frame, err error) {
 	channel := binary.BigEndian.Uint16(scratch[1:3])
 	size := binary.BigEndian.Uint32(scratch[3:7])
 
+	if r.maxFrameSize != nil {
+		if max := r.maxFrameSize.Load(); max > 0 && uint64(size)+frameHeaderSize > uint64(max) {
+			return nil, ErrFrameTooLarge
+		}
+	}
+
 	switch typ {
 	case frameMethod:
 		if frame, err = r.parseMethodFrame(channel, size); err != nil {
--- a/read.go
+++ b/read.go
@@ -55,7 +55,10 @@ func (r *reader) ReadFrame() (frame frame, err error) {
 	size := binary.BigEndian.Uint32(scratch[3:7])
 
 	if r.maxFrameSize != nil {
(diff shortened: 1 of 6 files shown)


MirahImage (approved):

read.go
```diff
@@ -54,6 +54,12 @@ func (r *reader) ReadFrame() (frame frame, err error) {
 	channel := binary.BigEndian.Uint16(scratch[1:3])
 	size := binary.BigEndian.Uint32(scratch[3:7])
 
+	if r.maxFrameSize != nil {
+		if max := r.maxFrameSize.Load(); max > 0 && uint64(size)+frameHeaderSize > uint64(max) {
```
Zerpet: I'm a bit uneasy with the integer conversion. I understand this is to avoid int32 overflow. However, since we know that max frame size must account for the header, and we know that the header is always 7 bytes, why don't we rewrite as `size - 8 > max` (7 bytes from header + 1 byte from frame-end byte).

I'm ok with introducing a constant and rewrite the if as `size - frameHeaderSizePlusFrameEnd`

Zerpet (changes requested):

read.go
suchitd: I think you meant `size > (max-frameHeaderSize)` where [frameHeaderSize=8](https://github.com/rabbitmq/amqp091-go/blob/main/channel.go#L24) . 

With this, we don't need integer conversion. Let me make that change. Thanks!

Zerpet: You are right! Never review after 5pm 😄

Zerpet (approved):

PR #372, opened 2026-07-27
fix: close goroutine leaks from Close()/Reconnect() races

suchitd:
## Proposed Changes

Problems observed:

1. Recovery already in progress when Close() is called: Close() would inspect connection/channel state without waiting for the in-flight Reconnect()/reconnectChannel() to settle, so it could tear down a stale snapshot while Reconnect() was still redialing, reopening channels, or spawning a new reader/heartbeater — resources Reconnect() built after Close() had already returned to the caller. Result: the connection's reader/heartbeater and any open channels' watchChannel/consumers goroutines leaked forever.

2. Close() winning that same race before Reconnect() even started: Close() could see IsClosed() == true on a connection that was merely mid-drop (not yet reconnected or exhausted) and return ErrClosed without any teardown, releasing the lock. The not-yet-started Reconnect() had no way to learn a close was already requested. Once closeInit was set, MaxRetryCount() returned 0 (it depends on IsRecoveryEnabled()), so Reconnect()'s retry loop never ran and it returned nil instead of an error. DefaultConnectionRecovery.OnConnectionClose only calls cleanup() on a non-nil error, so that silent nil skipped teardown entirely, orphaning everything forever.

Fixes:

1. Close()/CloseDeadline()/Channel.Close() now acquire c.reconnecting/ch.reconnecting — the same mutex Reconnect()/ reconnectChannel() hold for their entire duration — before inspecting state, so they always see the fully settled outcome (open, or fully torn down) rather than a mid-flight one.

2. Mark closeInit before checking IsClosed() or racing for c.reconnecting, and have Reconnect() re-check IsRecoveryEnabled() immediately after acquiring c.reconnecting, aborting with ErrClosed so it correctly routes through cleanup() instead of returning a silent nil.

Add TestReconnectAbortsWhenCloseWonTheRace, which forces interleaving (2) deterministically and asserts on Reconnect() never reaching its retry loop (no dial attempt, no StateReconnecting transition) rather than inferring it from timing. Interleaving (1) was verified the same way during development (forcing a busy dialer to prove Close() blocks for the in-flight attempt's full duration, then that cleanup() tears down cleanly), though that test isn't kept as a permanent regression test since it duplicates coverage already exercised by the recovery integration tests below.

Also update the recovery integration tests that use throwaway verification channels for expected-to-fail passive declares: running them on the recovering connection triggered the library's own per-channel auto-recovery (watchChannel -> OnChannelClose -> Channel.Reconnect) for a channel being intentionally broken, racing with the test's own teardown — an instance of problem (1) above. Use a separate non-recovering connection for these checks instead.

## Types of Changes

- [x] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

### DIFF
--- a/channel.go
+++ b/channel.go
@@ -2488,7 +2488,6 @@ func (ch *Channel) setupChannelBasic() error {
 func (ch *Channel) resetState() {
 	ch.closed.Store(false)
 	ch.destructed = false
-	ch.closeInit.Store(false)
 
 	ch.errors = make(chan *Error, 1)
 	ch.close = make(chan struct{})
--- a/channel.go
+++ b/channel.go
@@ -2343,6 +2343,15 @@ func (ch *Channel) Reconnect() error {
 // whether to send a channel.close courtesy frame to the broker before the next
 // retry (meaningful only when open succeeded but setup then failed).
 func (ch *Channel) openChannelSession() (openSucceeded bool, err error) {
+	// Close() may have marked closeInit and be racing us for ch.reconnecting;
+	// both callers hold that mutex, so this check is the single point that
+	// keeps either from resurrecting a channel Close() already committed to
+	// closing. See reconnectChannel() for the analogous per-caller check this
+	// supersedes.
+	if ch.closeInit.Load() {
+		return false, ErrClosed
+	}
+
 	// 1. Reset client-side state
 	ch.destructorM.Lock()
 	ch.m.Lock()
@@ -2411,9 +2420,11 @@ func (ch *Channel) reconnectChannel() error {
 	// ch.reconnecting between our check above and this point — in
 	// particular, its earlier closeRecovery() call runs before we've
 	// registered a cancel channel below, so that signal alone would
-	// otherwise be lost. Without this, a reconnectChannel() that wins the
-	// race would run its full retry loop while Close() waits behind
-	// ch.reconnecting for no reason. See Connection.Reconnect() for the
+	// otherwise be lost. This is a fast-path optimization on top of the
+	// authoritative check in openChannelSession(): without it, a
+	// reconnectChannel() that wins the race would still run its full retry
+	// loop (with backoff) before openChannelSession() rejects each attempt,
+	// instead of returning immediately. See Connection.Reconnect() for the
 	// analogous connection-level race.
 	if ch.closeInit.Load() {
 		return ErrClosed
--- a/channel.go
+++ b/channel.go
@@ -114,6 +114,8 @@ type Channel struct {
 	reconnecting sync.Mutex // Mutex for reconnecting channel.
 	lifeCycle    *lifeCycle // The current state of the channel.
 
+	closeInit atomic.Bool // Set by Close() before racing for reconnecting; see reconnectChannel().
+
 	// recoveringTopology is true while a recoverConnectionTopology pass owns
 	// this channel's reopen/redeclare sequence. watchChannel's NotifyClose
 	// listener checks this to avoid starting a redundant, competing
@@ -688,6 +690,16 @@ It is safe to call this method multiple times.
 */
 func (ch *Channel) Close() error {
 	ch.closeRecovery() // Stop any active recovery process
+
+	// Mark close-intent before racing for ch.reconnecting below. If
+	// reconnectChannel() hasn't started yet (e.g. it hasn't reached
+	// ch.reconnecting.Lock() and so closeRecovery() above found no cancel
+	// channel to close), it has no other way to learn a close was
+	// requested. reconnectChannel() re-checks this immediately after
+	// acquiring ch.reconnecting, so setting it first guarantees it aborts
+	// instead of running its full retry loop while this call waits behind
+	// it. See Connection.Close() for the analogous connection-level fix.
+	ch.closeInit.Store(true)
 
 	// Wait for any in-flight reconnectChannel() to fully settle before
 	// inspecting/tearing down state. See Connection.Close() for the analogous
@@ -2395,6 +2407,18 @@ func (ch *Channel) reconnectChannel() error {
 	ch.reconnecting.Lock()
 	defer ch.reconnecting.Unlock()
 
+	// Re-check: Close() may have marked closeInit and be racing us for
+	// ch.reconnecting between our check above and this point — in
+	// particular, its earlier closeRecovery() call runs before we've
+	// registered a cancel channel below, so that signal alone would
+	// otherwise be lost. Without this, a reconnectChannel() that wins the
+	// race would run its full retry loop while Close() waits behind
+	// ch.reconnecting for no reason. See Connection.Reconnect() for the
+	// analogous connection-level race.
+	if ch.closeInit.Load() {
+		return ErrClosed
+	}
+
 	if !ch.IsClosed() {
 		return nil
 	}
@@ -2481,6 +2505,7 @@ func (ch *Channel) setupChannelBasic() error {
 func (ch *Channel) resetState() {
 	ch.closed.Store(false)
 	ch.destructed = false
+	ch.closeInit.Store(false)
 
 	ch.errors = make(chan *Error, 1)
 	ch.close = make(chan struct{})
--- a/channel.go
+++ b/channel.go
@@ -689,6 +689,12 @@ It is safe to call this method multiple times.
 func (ch *Channel) Close() error {
 	ch.closeRecovery() // Stop any active recovery process
 
+	// Wait for any in-flight reconnectChannel() to fully settle before
+	// inspecting/tearing down state. See Connection.Close() for the analogous
+	// connection-level race this prevents.
+	ch.reconnecting.Lock()
+	defer ch.reconnecting.Unlock()
+
 	if ch.IsClosed() {
 		return nil
 	}
--- a/channel.go
+++ b/channel.go
@@ -63,7 +63,7 @@ should be discarded and a new channel established.
 type Channel struct {
 	destructorM sync.Mutex   // Mutex for destroying the channel.
 	destructed  bool         // Will be true if the channel has been destroyed, false otherwise.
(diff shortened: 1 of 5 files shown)


channel.go
```diff
@@ -114,6 +114,8 @@ type Channel struct {
 	reconnecting sync.Mutex // Mutex for reconnecting channel.
 	lifeCycle    *lifeCycle // The current state of the channel.
 
+	closeInit atomic.Bool // Set by Close() before racing for reconnecting; see reconnectChannel().
+
```
Zerpet, line 118: We have the following bool flags before this change:

1. `destructed`
2. `cleanedUp`
3. `closed`

Along with other 4 mutexes. This PR introduces a 4th bool flag.

I understand all of these are necessary one way or another to cover specific cases. However, instead of introducing yet another bool flag, I would like to consider an internal state field as an enum, and simplify this already bloated struct. Ideally it should not be an atomic, I believe this case can be successfully covered by a mutex. It can be somewhat tricky because we have 4 mutexes already, and it's nearly impossible to reason about mutex usage along any operation.

suchitd: I do agree with you, however this flag is analogous to the one in connection: https://github.com/rabbitmq/amqp091-go/blob/main/connection.go#L149

Its purpose is to mark that the Close method has been called and no other flows should resurrect the channel by any means leaving leaked goroutines. Its the same logic followed as in connection recovery.

> I would like to consider an internal state field as an enum

Let me see what we can do with existing lifecycle state. I can't promise anything yet :)

suchitd: The existing boolean flags `destructed `, `closed` are solving different synchronisation needs that too [avoiding lock contentions](https://github.com/rabbitmq/amqp091-go/blob/main/CLAUDE.md?plain=1#L108).
We want similar latch that will notify(without contending for lock) to on-going recovery to bail out immediately so that Close() acquires ch.reconnecting lock without waiting to complete full recovery retry loop. 
We can't use lifecycle state as its behind reconnecting lock and bringing it outside will introduce real regressions (it has its own listener notifications with a terminal state as Closed. It can't go back from Closed to any other state). So, thats completely out of question.

I asked Claude to evaluate a proposed internal state machine that has enums per boolean flag, and it listed more negatives than positives. Based on the report, we may be able to convert `destructed` and `closeInit` into enum with some additional logic of keeping closeInit as terminal state, but I felt its not worth it.

I think having `closeInit atomic.bool` is the cheapest and most safe option we have( at least w.r.t. this PR).

Also, I changed `cleanedUp atomic.Bool` --> `closeOnce sync.Once` to solve a rare double close of consumer channel issue and simplified Connection/Channel teardown paths.

Zerpet:
> 🤖 **AI-generated review** (Claude Code), posted with a human in the loop reviewing before submission.

## Overview

Fixes two races between `Close()`/`CloseDeadline()`/`closeWith()` and the in-flight `Reconnect()`/`reconnectChannel()` paths:

1. **Close() racing past a live Reconnect()**: `Close()` could inspect a stale `IsClosed()` snapshot while `Reconnect()` was mid-flight (redialing, reopening channels, spawning a new reader/heartbeater), tearing down state out from under it and orphaning the newly built goroutines.
2. **Close() winning before Reconnect() starts**: `Close()` could set `closeInit`, see `IsClosed() == true`, and return `ErrClosed` without any teardown — leaving a not-yet-started `Reconnect()` to fall through as a silent no-op (`MaxRetryCount()` → 0 once `closeInit` is set), which skips `cleanup()` since `OnConnectionClose` only calls it on a non-nil error.

Fix strategy: `Close()`/`CloseDeadline()`/`closeWith()`/`Channel.Close()` now mark close-intent *before* acquiring `reconnecting` (the same mutex `Reconnect()`/`reconnectChannel()` hold for their full duration), then wait on it before inspecting state. `Reconnect()`/`reconnectChannel()` re-check close-intent immediately after acquiring `reconnecting` and abort with `ErrClosed` if it was set. Also refactors `shutdown()`/`cleanup()` teardown into a shared, idempotent `closeResources()` (via `sync.Once` for `Channel`, naturally idempotent nil-outs for `Connection`), and switches several `recovery_test.go` verification checks off the recovering connection onto a dedicated non-recovering connection to avoid triggering the library's own auto-recovery mid-test.

## Verification performed

- Built and `go vet`'d the PR branch cleanly.
- Ran the full non-integration unit suite with `-race`: pass.
- Ran the two new race-interleaving tests with `-race -count=5`: stable, no flakes.
- Traced lock ordering for the new `reconnecting`-then-`m`/`notifyM`/`destructorM` paths in `Close()`, `closeChannel()`, `shutdown()`, `closeResources()` — found no new deadlock (defers unwind in the right order, `closeChannel`/`shutdown` never re-acquire `reconnecting`).

## Strengths

- The core fix is well targeted: mirroring the "mark intent → acquire the same mutex the other side holds → re-check intent" pattern symmetrically on both connection and channel levels is the right shape for this class of race.
- `TestReconnectAbortsWhenCloseWonTheRace` / `TestReconnectChannelAbortsWhenCloseWonTheRace` deterministically force the exact interleaving (hold `reconnecting` externally, let `Reconnect()` pass its first check and block, then flip `closeInit`, then release) rather than relying on timing/sleeps to hope the race occurs — this is exactly right for regression-testing a race.
- `closeResources()` consolidation removes real duplication between `shutdown()` and `cleanup()`, and using `sync.Once` for the channel case is a clean way to guarantee single-teardown regardless of which path reaches it first, replacing the previous `atomic.Bool` CAS-guard that did the same thing.
- Comments are unusually thorough and cross-reference the connection/channel analogs, which will help future maintainers.

## Issues / risks

**1. Primary race (in-flight Reconnect torn down) has no permanent regression test.**
The PR body states this was "verified during development" but not kept as a permanent test because it "duplicates coverage already exercised by the recovery integration tests." Only the second race (Close winning before Reconnect starts) got a dedicated deterministic unit test. Integration tests are timing-sensitive and don't reliably force this specific interleaving (Close() blocking on `reconnecting` for an in-flight Reconnect's full duration, then finding an already-open connection/channel that Reconnect() rebuilt). Given this is literally problem #1 in the PR description and the harder-to-hit race, it'd be worth keeping a deterministic unit test for it too — the same "hold `reconnecting` externally, release it later" technique used for the other two tests would work here.

**2. `CloseDeadline`'s deadline no longer bounds the full call.**
The PR's own added doc comment admits this: "If an in-flight Reconnect() is holding `c.reconnecting`, this call blocks behind it first — that wait is not covered by deadline and can run for the duration of Reconnect()'s full retry loop." This is a real behavior change for callers who pass a tight deadline expecting bounded latency (e.g., a shutdown path with an SLA) — they can now block for up to `MaxRetryCount * RetryInterval` regardless of the deadline argument. Worth considering whether `CloseDeadline` should race the `reconnecting.Lock()` against the deadline (e.g. a `TryLock`-in-a-loop or a lock-with-timeout helper) rather than just documenting the gap.

**3. `closeWith()` gained the same `closeRecovery()` + wait-on-`reconnecting` treatment, but it's not mentioned in the PR description and has no dedicated test.**
`closeWith()` is invoked asynchronously (`go func(){ closeWith(...) }()`) from `dispatch0`/`dispatchClosed` for protocol-violation cases (unexpected frame / unexpected method on a closed channel). Previously it neither called `closeRecovery()` nor waited on `reconnecting`. Since it's async this doesn't block the reader, but it introduces the same theoretical hazard the other three entry points have: after waiting out a Reconnect(), it re-checks `IsClosed()` against what may now be a freshly reconnected connection (new socket, new reader) and, if that check passes, will send a `connection.close` and shut down the just-recovered connection based on a stale error from the old socket. This narrow window likely predates the PR in a smaller form, but the change measurably widens it (adds a real wait where there was none) without a comment explaining why `closeWith` needed the same treatment or a test covering it specifically. Worth a one-line justification at minimum.

**4. Minor inconsistency: `Channel.closeInit` is `atomic.Bool`, `Connection.closeInit` is a plain `bool` guarded by `closeM`.**
Not a bug — both are used correctly and consistently in their own contexts — but it's a small asymmetry between two structurally identical mechanisms in the same PR. Unifying the approach would make the "connection-level fix ↔ channel-level fix" cross-references in the comments literally as well as conceptually parallel.

**5. Repeated three-way duplication** of the "mark closeInit, check initiated, lock reconnecting, check IsClosed" prologue across `Close()`, `CloseDeadline()`, and `closeWith()`. A small private helper (e.g. `c.beginClose() (initiated bool)` that also acquires/returns an unlock-func for `reconnecting`) could collapse this, but it's a nice-to-have, not a blocker.

## Test-file changes (recovery_test.go)

The switch to dedicated non-recovering connections for passive-declare verification checks is a reasonable and well-explained fix for the described test-vs-auto-recovery race. One nit: `TestConnectionRecoveryTopologyDisabled` now opens `freshConn` earlier than before and reuses it for three separate checks — that's fine, just note the connection's lifetime now spans more of the test than the comment right above the first use suggests at a glance.

## Summary

Correct, well-reasoned fix for a real (and nasty) class of resource leak, with solid targeted regression tests for one of the two races and passing `-race` runs locally. Before merge: a deterministic regression test for race #1 (not just "verified during development"), and either a fix or an explicit accepted-tradeoff note for `CloseDeadline`'s now-unbounded wait, since that changes an exported method's documented contract.

suchitd:
Thanks Aitor!

Issues/Risks reported by AI review:
1. Fixed. Added dedicated unit test.
2. Its not a breaking change, and even the behaviour is same for older clients as it won't have recovery enabled. If recovery is in progress and at that exact moment closeDeadline() is called, it will wait for reconnecting lock(we need this to handle the exact race issue this PR is for) and the same is documented. 
3. `closeWith` is called when the protocol violation occurs, in that case it is expected that the connection is closed.
4. Connection closeInit can't be converted to atomic.bool because recovery needs to hold closeM until it swaps the new connection.
5. Minor, I'll handle this as a follow-up PR.

PR #377, opened 2026-07-31
Enforce frame size in pre-negotiation state

Zerpet:
As per AMQP 091 spec, before Tune frames are exchanged, the max frame size must be 4096 bytes.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -381,9 +381,12 @@ func Open(conn io.ReadWriteCloser, config Config) (*Connection, error) {
 		errors:                make(chan *Error, 1),
 		close:                 make(chan struct{}),
 		deadlines:             make(chan readDeadliner, 1),
-		Config:                config,
-		lifeCycle:             newLifeCycle(),
+		// TODO: Connection has Config and also an atomic int for MaxFrameSize. Duplication to simplify.
+		Config:    config,
+		lifeCycle: newLifeCycle(),
 	}
+	// Before max frame size is negotiated in Tune, the spec sets a ceiling of 4096 bytes
+	c.maxFrameSize.Store(frameMinSize)
 	go c.reader(conn)
 	err := c.open(config)
 	if err == nil {
@@ -1298,7 +1301,7 @@ func (c *Connection) openTune(config Config, auth Authentication) error {
 	// minimum floor of frameMinSize (4096 bytes) to prevent malicious servers
 	// from forcing extreme fragmentation and CPU overhead.
 	c.Config.FrameSize = negotiateFrameSize(config.FrameSize, int(tune.FrameMax))
-	// This is the only place maxFrameSize is written. reader.ReadFrame relies on
+	// reader.ReadFrame relies on
 	// any nonzero value here being >= frameMinSize (negotiateFrameSize's floor)
 	// to safely subtract frameHeaderSize without underflow — keep it that way if
 	// another write path is ever added.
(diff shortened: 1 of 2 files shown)


connection.go
```diff
@@ -381,9 +381,12 @@ func Open(conn io.ReadWriteCloser, config Config) (*Connection, error) {
...
 		close:                 make(chan struct{}),
 		deadlines:             make(chan readDeadliner, 1),
-		Config:                config,
-		lifeCycle:             newLifeCycle(),
+		// TODO: Connection has Config and also an atomic int for MaxFrameSize. Duplication to simplify.
+		Config:    config,
+		lifeCycle: newLifeCycle(),
 	}
+	// Before max frame size is negotiated in Tune, the spec sets a ceiling of 4096 bytes
+	c.maxFrameSize.Store(frameMinSize)
```
suchitd, line 389: You may want to update [this comment](https://github.com/rabbitmq/amqp091-go/blob/main/connection.go#L1301) as it no longer remain the only place that updates `c.maxFrameSize`

connection.go
```diff
@@ -381,9 +381,12 @@ func Open(conn io.ReadWriteCloser, config Config) (*Connection, error) {
 		errors:                make(chan *Error, 1),
 		close:                 make(chan struct{}),
 		deadlines:             make(chan readDeadliner, 1),
-		Config:                config,
-		lifeCycle:             newLifeCycle(),
+		// TODO: Connection has Config and also an atomic int for MaxFrameSize. Duplication to simplify.
```
suchitd, line 384: The atomic `c.maxFrameSize` is intended to avoid data in `reader`.

connection_unit_test.go
```diff
@@ -156,6 +156,75 @@ func TestReconnectRestoresSASLCredentials(t *testing.T) {
...
+	if got != uint32(frameMinSize) {
+		t.Errorf("maxFrameSize = %d, want %d (frameMinSize)", got, frameMinSize)
+	}
+}
+
+// TestOpenMaxFrameSizeIsNotZero verifies that after Open() initializes the connection,
+// maxFrameSize has been set to a non-zero value. This guards against a regression
+// where maxFrameSize initialization could be accidentally removed, leaving it at
+// the default zero value and allowing oversized frame allocations before negotiation.
+func TestOpenMaxFrameSizeIsNotZero(t *testing.T) {
```
suchitd, line 198: `c.maxFrameSize.Store(frameMinSize)` is stored unconditionally, regardless of what `Config` has. I think `TestOpenInitializesMaxFrameSize` is sufficient to exercise and validate the new code path.

connection.go
Zerpet: Then why we keep both? 
An entire `Config`, where we don't use most fields, and the atomic. We may be able to drop the `Config` and keep only relevant fields. I believe the comment is still worth keeping to revisit and simplify the struct in the future

suchitd: Config is exported struct and changing/removing the field would be a breaking change. 
We can revisit this in `2.0.0`.

connection_unit_test.go
Zerpet: This test verifies that the value in `Config` is not considered during Connection-Start and Tune. I understand that the atomic is set unconditionally in the current code. This test verifies that future changes don't do the wrong thing. Even if it doesn't provide additional coverage, it still provides value.

suchitd (approved):

PR #389, opened 2026-08-31
fix: preserve caller-provided SASL credentials across Reconnect()

suchitd:
## Proposed Changes

Reconnect() unconditionally reset Config.SASL to nil and rebuilt it from the connection URL on every retry, discarding whatever Authentication the caller originally configured via DialConfig. This silently replaced custom Authentication implementations (OAuth2, Kerberos, etc.) with a URL-derived PlainAuth, and broke PlainAuth/AMQPlainAuth credentials supplied out-of-band that differed from the URL.

Building on the #388 clone fix (which stops openStart from mutating the caller's own Authentication objects), Connection now retains originalSASL, a clone of Config.SASL captured once at initial Open, before it's narrowed to the chosen mechanism and before openComplete zeroes its credentials. Reconnect() restores Config.SASL from that clone, falling back to deriving credentials from the URL only when none were ever configured.

## Types of Changes

- [x] Bugfix 
- [ ] New feature 
- [ ] Breaking change 
- [ ] Documentation 
- [ ] Cosmetics 

## Checklist

### DIFF
--- a/auth.go
+++ b/auth.go
@@ -90,9 +90,6 @@ func pickSASLMechanism(client []Authentication, serverMechanisms []string) (auth
 	return
 }
 
-// cloneAuthentications returns a clone of each entry in auths, preserving
-// order. Used to retain an independent copy of caller-provided
-// Authentication candidates that survives credential zeroing.
 func cloneAuthentications(auths []Authentication) []Authentication {
 	if auths == nil {
 		return nil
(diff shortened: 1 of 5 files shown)


auth.go
```diff
@@ -90,7 +90,37 @@ func pickSASLMechanism(client []Authentication, serverMechanisms []string) (auth
...
+// aliasing it here introduces no corruption on this package's side. It
+// does mean a custom Authentication reused across reconnects is the
+// caller's literal object: if the caller mutates its internal state in
+// place (e.g. rotating a token), that mutation is visible on the next
+// reconnect too, since it's the same instance, not a copy. This makes a
+// custom Authentication the right choice when secrets/passwords rotate over
+// the connection's lifetime (e.g. via UpdateSecret): unlike PlainAuth/
+// AMQPlainAuth, which are frozen as a deep copy at Open time, a custom
+// Authentication that reads its secret dynamically in Response() will
+// always have Reconnect() use the latest value.
```
Zerpet: Do we need a 16-line comment for a 10-line function? 😄

auth.go
```diff
@@ -90,7 +90,37 @@ func pickSASLMechanism(client []Authentication, serverMechanisms []string) (auth
 	return
 }
 
-// cloneAuthentication returns a clone of auth.
+// cloneAuthentications returns a clone of each entry in auths, preserving
+// order. Used to retain an independent copy of caller-provided
+// Authentication candidates that survives credential zeroing.
```
Zerpet: I would also remove these comments, I think the function is simple and clear and it doesn't need a comment. The function is also unexported, so it doesn't require a go-doc comment.

recovery_test.go
```diff
@@ -296,6 +296,161 @@ func TestConnectionRecoveryConsume(t *testing.T) {
...
+		}
+		if string(d.Body) != preRecoveryMessage {
+			t.Fatalf("Expected message '%s', got: %s", preRecoveryMessage, string(d.Body))
+		}
+		t.Logf("Received message pre-recovery: %s", string(d.Body))
+	case <-time.After(5 * time.Second):
+		t.Fatalf("Timeout waiting for receive message pre-recovery: %s", preRecoveryMessage)
+	}
+
+	// Register with connection for NotifyStateChange
```
Zerpet:
```suggestion
```

recovery_test.go
```diff
@@ -296,6 +296,161 @@ func TestConnectionRecoveryConsume(t *testing.T) {
...
+	stateChanged := make(chan *StateChanged, 10)
+	conn.NotifyStateChange(stateChanged)
+
+	// Register with channel for NotifyStateChange
```
Zerpet:
```suggestion
```

recovery_test.go
```diff
@@ -296,6 +296,161 @@ func TestConnectionRecoveryConsume(t *testing.T) {
...
+	chanStateChanged := make(chan *StateChanged, 10)
+	ch.NotifyStateChange(chanStateChanged)
+
+	// Call Http API to close the current connection
```
Zerpet:
```suggestion
```

Zerpet (approved):
I'm squirming with the new addition of a field to duplicate even more information, but I don't see an alternative without breaking changes 😮‍💨 Chop some/most comments and you have my approval ✅

auth.go
suchitd: Removed

auth.go
suchitd: Trimmed the comment. Thanks!

PR #396, opened 2026-09-19
docs: document URI credential percent-encoding requirement (#305)

cipherprofessor:
## Summary

Closes the documentation gap raised in #305: a username or password containing a reserved URI character (e.g. a literal `?`) breaks `ParseURI`/`Dial` with a confusing stdlib error, and nothing in the docs said credentials need to be percent-encoded.

As pointed out in that issue, `ParseURI` parses with `net/url`, so per RFC 3986 §3.2.1 any reserved character in the username or password (`:`, `@`, `/`, `?`, `#`, `%`, ...) must be percent-encoded. That's expected behavior given the stdlib parser, but it wasn't written down anywhere, and the resulting error (e.g. `invalid port ":password" after host` for a `?` in the password) gives no hint about the real cause.

## Changes

- `ParseURI`'s doc comment now states the percent-encoding requirement and shows a safe way to build such a URI with `net/url.UserPassword` (which encodes correctly).
- `Dial`'s doc comment points to `ParseURI` for this.
- `ParseURI`'s `url.Parse` error is wrapped (via `%w`, so `errors.Is`/`errors.As` on the underlying error still works) with a pointer to the same fix, so hitting this actually leads somewhere instead of a bare `net/url` error.

No parsing behavior changes — this is documentation plus a friendlier error message.

## Test plan
- [x] `gofmt -l .` — clean
- [x] `go build ./...` — succeeds
- [x] `go vet ./...` — clean
- [x] `go test -race ./...` — all unit tests pass
- [x] Verified the reported repro (`amqp://cert_refresh:password?@192.168.0.9:5672/`) now fails with a message that includes the percent-encoding pointer instead of a bare `invalid port` error
- [x] Verified the doc comment's own suggested fix (`url.UserPassword("cert_refresh", "password?")`) round-trips correctly through `ParseURI`


Zerpet (changes requested):
Content looks good. Blocking on CLA.

Please note that since the 28th of July 2025, all external contributors to RabbitMQ will have to sign a CLA again.

You have detailed instructions in [1][2].

Note companies that contribute to RabbitMQ only need to sign the CLA once, there is no need to make every
contributing employee to do so separately.

https://github.com/rabbitmq/cla
https://github.com/rabbitmq/rabbitmq-server/blob/main/CONTRIBUTING.md

cipherprofessor:
CLA signed. Let me know if there's anything else needed to move this forward — happy to make any further changes if useful.

Zerpet (approved):

